PR #27, opened 2017-01-07
The printing function is named write_str

samueltardieu:


japaric:
`write_fmt` is the right method here.

The [`fmt::Write`](https://doc.rust-lang.org/core/fmt/trait.Write.html) trait has three methods:

``` rust
pub trait Write {
    fn write_str(&mut self, s: &str) -> Result;

    fn write_char(&mut self, c: char) -> Result { ... }
    fn write_fmt(&mut self, args: Arguments) -> Result { ... }
}
```

We have to use `write_fmt` inside the macro because `format_args!` returns a value with type `Arguments` so `write_str` wouldn't work because of the signature.

The confusing part is that we are *not* implementing the `write_fmt` method for `Serial`! Instead, we only implement `write_str`. The thing is that `write_char` and `write_fmt` are default methods so once we implement `write_str` for a type, the compile will automatically implement the other two for the type.

Does that make sense?

samueltardieu:
Sure. Thank you for the explanation.

PR #50, opened 2017-10-09
Fixes typo.

cldershem:
Added a missing word. It could be 'you' or 'we', I went with 'we' to make us all feel a little less alone in this world.


japaric:
Thanks, @cldershem.

PR #54, opened 2017-10-10
Updated page number

cldershem:
Page number pointed to the end of the section instead of the beginning of the section.


japaric:
Thanks, @cldershem.

PR #67, opened 2018-02-28
add symlinks to /target

### LINKED ISSUE #66: Target directory not clear in 05-led-roulette/build-it
When executing the following from the `discovery/src/05-led-roulette` directory:

```xargo build --target thumbv7em-none-eabihf```

It's not clear the the resulting target is `discovery/target` since the previous page says "Now, jump into the `src/05-led-roulette directory`."

japaric:
moving to workspaces means that now the target directory will now be at the root of the repo. This
doesn't match the text which says that target is right next to the Cargo.toml of each Cargo project.
Workaround (only on *nix?) the problem by adding symlinks

closes #66

cc @WimbledonLabs thoughts?


WimbledonLabs:
Yeah, that makes sense. To be honest I haven't written much rust, so I don't know another way to organize the workspace.

japaric:
Alright, let's go ahead with this.

bors r+

> I don't know another way to organize the workspace.

I don't know either as I don't usually use workspaces but the whole point of workspaces is to have a single target directory shared between different Cargo projects.

PR #74, opened 2018-03-24
Fix in chapter 9: set OPM and clear CEN (for one pulse mode)

pheki:
First of all, thanks for your great job in the world of programming microcontrollers in rust!
I believe the bit set and clear were mixed up. The registers are explained in the page 678 of the user manual.
Thanks!


japaric:
Thanks @pheki!

PR #84, opened 2018-05-26
Point out source of "duplicate lang item"

chrysn:
If a reader does not carefully follow the previous page's steps (the "Jump into" step is meant as a `cd`, not just browse there), a reader might (ok, I did) execute the `cargo build` in the discovery directory.

With the error message that ensues (``duplicate lang item found: `panic_fmt`.``), it make take some time for the reader to find the mistake (mine: 5min); a cautionary note might save others that time.


japaric:
Thanks for the PR. I have added a comment before the `cargo build` command. (The error you get now is different as `panic_fmt` no longer exists).

PR #96, opened 2018-07-21
Highlight the fact that openocd and itmdump must be run in the same d…

### LINKED ISSUE #32: On Fedora 24 itm.txt file location is not correct
From https://japaric.github.io/discovery/06-hello-world/README.html The following instructions appear to be wrong

```
# *nix
$ cd /tmp

# Windows
$ cd %TEMP%

$ itmdump itm.txt
```

Instead you must run `itmdump itm.txt` in the directory you ran gdb in with your current instructions. I believe you can make them correct by changing the following:

```
(gdb) monitor tpiu config internal itm.txt uart off 8000000
```

To

```
(gdb) monitor tpiu config internal /tmp/itm.txt uart off 8000000
```

johnthagen:
...irectory

Fixes #32


japaric:
LGTM, thanks!

PR #97, opened 2018-07-22
Add missing words

MajorBreakfast:


src/05-led-roulette/build-it.md
```diff
@@ -22,8 +22,8 @@ target. That's done using `rustup`:
 $ rustup target add thumbv7em-none-eabihf
 ```
 
-You only the above step once; `rustup` will re-install a new standard library (`rust-std` component)
-whenever you update your toolchain.
+You only need to do the above step once; `rustup` will re-install a new standard library
+(`rust-std` component) whenever you update your toolchain.
```
MajorBreakfast: That's not new: I just moved the line-break

japaric:
Thanks @MajorBreakfast !

PR #100, opened 2018-07-31
Fix LED Compass project GDB breakpoint

johnthagen:
This fix causes the project to break at the normal start of the user's `main` function as it does in all of the other examples, allowing `layout src` to immediately work.


japaric:
LGTM

PR #108, opened 2018-08-21
install a newer gcc

japaric:
should fix the build

r? @rust-embedded/resources
cc @andre-richter

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,9 @@
 language: rust
 rust: nightly
 
-addons:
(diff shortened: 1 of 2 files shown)


japaric:
@andre-richter bors only reads normal comments; it doesn't read review comments so you always need an extra comment. (Annoying, I know)

bors r=andre-richter

japaric:
Oh wait, it worked with the review comment. Maybe they fixed it.

PR #110, opened 2018-08-28
use LLD, suppress warnings, no GCC, no aborts, fix links

japaric:
see individual commit messages for further details

r? @rust-embedded/resources (anyone)

### DIFF
--- a/src/03-setup/linux.md
+++ b/src/03-setup/linux.md
@@ -124,7 +124,7 @@ $ #  ^^^^
 
 If `uucp` appears in the output. You are all set! Go to the [next section]. Otherwise, keep reading:
 
-[next section]: 03-setup/verify.html
+[next section]: /03-setup/verify.html
 
 - Add yourself to the `uucp` group.
 
--- a/src/03-setup/linux.md
+++ b/src/03-setup/linux.md
@@ -8,7 +8,6 @@ Here are the installation commands for a few Linux distributions.
 
 ``` console
 $ sudo apt-get install \
-  gcc-arm-none-eabi \
   gdb-arm-none-eabi \
   minicom \
   openocd
@@ -18,7 +17,6 @@ $ sudo apt-get install \
 
 ``` console
 $ sudo dnf install \
-  arm-none-eabi-gcc-cs \
   arm-none-eabi-gdb \
   minicom \
   openocd
@@ -28,22 +26,25 @@ $ sudo dnf install \
 
 ``` console
 $ sudo pacman -S \
-  arm-none-eabi-gcc \
   arm-none-eabi-gdb \
   minicom \
   openocd
 ```
 
-- Other distros
+- `arm-none-eabi-gdb` for other distros
 
-For distros that don't have packages for [ARM's pre-built toolchain](https://developer.arm.com/open-source/gnu-toolchain/gnu-rm/downloads), download the "Linux 64-bit" file and put its `bin` directory on your path. Here's one way to do it:
+For distros that don't have packages for [ARM's pre-built
+toolchain](https://developer.arm.com/open-source/gnu-toolchain/gnu-rm/downloads),
+download the "Linux 64-bit" file and put its `bin` directory on your path.
+Here's one way to do it:
 
 ``` console
 $ mkdir -p ~/local && cd ~/local
 $ tar xjf /path/to/downloaded/file/gcc-arm-none-eabi-7-2017-q4-major-linux.tar.bz2.tbz
 ```
 
-Then, use your editor of choice to append to your `PATH` in the appropriate shell init file (e.g. `~/.zshrc` or `~/.bashrc`):
+Then, use your editor of choice to append to your `PATH` in the appropriate
+shell init file (e.g. `~/.zshrc` or `~/.bashrc`):
 
 ```
 PATH=$PATH:$HOME/local/gcc-arm-none-eabi-7-2017-q4-major/bin
--- a/src/03-setup/macos.md
+++ b/src/03-setup/macos.md
@@ -5,7 +5,6 @@ All the tools can be install using [Homebrew]:
 [Homebrew]: http://brew.sh/
 
 ``` console
-$ brew cask install gcc-arm-embedded
 $ brew install minicom openocd
 ```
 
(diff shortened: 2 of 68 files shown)


japaric:
ping @thejpster @jamesmunns @therealprof @andre-richter anyone up for reviewing this?

src/03-setup/linux.md
```diff
@@ -28,22 +26,25 @@ $ sudo dnf install \
...
   minicom \
   openocd
 ```
 
-- Other distros
+- `arm-none-eabi-gdb` for other distros
 
-For distros that don't have packages for [ARM's pre-built toolchain](https://developer.arm.com/open-source/gnu-toolchain/gnu-rm/downloads), download the "Linux 64-bit" file and put its `bin` directory on your path. Here's one way to do it:
+For distros that don't have packages for [ARM's pre-built
+toolchain](https://developer.arm.com/open-source/gnu-toolchain/gnu-rm/downloads),
```
jamesmunns: Do cross line tags cause problems?

src/03-setup/macos.md
```diff
@@ -5,7 +5,6 @@ All the tools can be install using [Homebrew]:
 [Homebrew]: http://brew.sh/
 
 ``` console
```
jamesmunns: Do you still need this for gdb?

src/05-led-roulette/.cargo/config
```diff
@@ -1,6 +1,6 @@
 [target.thumbv7em-none-eabihf]
 runner = "arm-none-eabi-gdb"
 rustflags = [
+  "-C", "linker=rust-lld",
```
jamesmunns: :tada:

src/08-leds-again/auxiliary/src/lib.rs
```diff
@@ -33,19 +35,25 @@ fn panic(info: &PanicInfo) -> ! {
 
     cortex_m::asm::bkpt();
 
-    loop {}
+    loop {
+        atomic::compiler_fence(Ordering::SeqCst)
```
jamesmunns: do these compiler fences affect the output binary size at all? I know they won't insert any code directly, but do they change the code generation by LLVM at all?

src/03-setup/macos.md
japaric: Yes, you are right; this is still needed for GDB. I'll add it back.

japaric:
@jamesmunns thanks for the review. I pushed a fix which invalidated your previous review so you'll have to re-review this.

src/08-leds-again/auxiliary/src/lib.rs
jamesmunns: Gotcha. I didn't know if this would prevent LLVM from optimizing something (causing a larger binary). Thanks for the clarification.

jamesmunns: Actually. Do you explain what these are/why they are necessary anywhere?

japaric: In the commit message, but I can add a comment like [this one](https://github.com/rust-embedded/cortex-m-rt/blob/795c86c296b36fdca75a742ab20ea678d71039af/src/lib.rs#L534) to each loop.

jamesmunns: I think just on the first instance would be fine IMO, or even in the text itself.

japaric: This code is not user facing. It's part of the auxiliary crate that expose some API that the reader will use, not part of the code they'll modify. 

I'll add a comment to each file since curious readers may look at the code.

src/05-led-roulette/my-solution.md
```diff
@@ -53,8 +53,9 @@ Binary size is something we should always keep an eye on! How big is your soluti
 that using the `size` command on the release binary:
 
 ``` console
-$ cargo size -- target/thumbv7em-none-eabihf/{debug,release}/led-roulette
-target/thumbv7em-none-eabihf/debug/led-roulette  :
+$ # equivalent to size target/thumbv7em-none-eabihf/debug/led-roulette
+$ cargo size --bin led-roulette -- -A
+led-roulette  :
```
Caemor: Leads to the following error for me:
```
cargo size --bin led-roulette -- -A
   Compiling led-roulette v0.1.0 (file:///home/caemor/git/discovery/src/05-led-roulette)                
error: language item required, but not found: `eh_personality`

error: aborting due to previous error

error: Could not compile `led-roulette`.

To learn more, run the command again with --verbose.
```
So I still needed to use 
`cargo size --bin led-roulette --target thumbv7em-none-eabihf -- -A`

cargo size (cargo-binutils) is installed in version 0.1.2.

src/05-led-roulette/my-solution.md
```diff
@@ -75,7 +76,8 @@ section               size        addr
 .debug_aranges         160         0x0
 Total               747766
 
-target/thumbv7em-none-eabihf/release/led-roulette  :
+$ cargo size --bin led-roulette -- -A
+led-roulette  :
```
Caemor: This is exactly the same command as above and shouldn't lead to different results (release vs debug size).
It looks like the 'release'-parameter is missing.

So I used nearly same as above but with added 'release' tag to get a result.
`cargo size --bin led-roulette --target thumbv7em-none-eabihf --release -- -A`

japaric:
@Caemor Good catch! Thanks.

re-r? @jamesmunns or anyone from @rust-embedded/resources. Please also `bors r+` this, otherwise it falls off from my radar and it takes me some time to notice it's approved but not merged.

src/03-setup/linux.md
```diff
@@ -28,22 +26,25 @@ $ sudo dnf install \
...
 ``` console
 $ sudo pacman -S \
-  arm-none-eabi-gcc \
   arm-none-eabi-gdb \
   minicom \
   openocd
 ```
 
-- Other distros
+- `arm-none-eabi-gdb` for other distros
```
therealprof: We should probably also mention the possibility of using a multi-architecture gdb

therealprof:
If `cargo-bloat` didn't just break it would have been nice to add some info on how to figure out where the flash is spent but I guess the point is moot now.

Let's land this.

bors r+

Issue #123, opened 2018-10-13
Relative links are missing bookshelf/discovery in URL


adamgreen:
Images aren't showing up in documentation. For example, if I navigate to https://rust-embedded.github.io/bookshelf/discovery/02-requirements/index.html none of the images linked from this page are rendered.

An example of a broken link found on this page:
https://rust-embedded.github.io/assets/f3.jpg

I think it should actually be:
https://rust-embedded.github.io/bookshelf/discovery/assets/f3.jpg

adamgreen:
It appears that PR #121 is attempting to fix the image links part of this issue.

ChouzArt:
Didn't notice the 404 on those relative links. Could be more but already made a new PR to fix those too

adamgreen:
This is a duplicate of https://github.com/rust-embedded/bookshelf/issues/4

PR #131, opened 2018-10-18
To fix 03-setup/Readme.md references to next section links

ChouzArt:
As seen on #123

### DIFF
--- a/src/03-setup/README.md
+++ b/src/03-setup/README.md
@@ -109,6 +109,6 @@ LLVM (http://llvm.org/):
 
 Now follow the instructions specific to the OS you are using:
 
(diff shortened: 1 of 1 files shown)


therealprof:
@rust-embedded/resources The markdown stuff is acting all funny. The absolute paths are different than what they're supposed to be and the relative paths don't work. Any ideas?

ChouzArt:
This is related to [rust-embedded/bookshelf#4](https://github.com/rust-embedded/bookshelf/issues/4).
Bookshelf version is broken, standalone show images.

It can be fixed doing this trick [rust-embedded/book#38](https://github.com/rust-embedded/book/pull/38).

In this case, inserting /discovery/ into the link

PR #132, opened 2018-10-22
Update .cargo/config diff baseline

adamgreen:
The diff in 06-hello-world/README.md for changes to be made to
.cargo/config are based off of an older version of the config file than
what is found in the repository currently.

### DIFF
--- a/src/05-led-roulette/.cargo/config
+++ b/src/05-led-roulette/.cargo/config
@@ -1,6 +1,5 @@
 [target.thumbv7em-none-eabihf]
 runner = "arm-none-eabi-gdb -q"
 rustflags = [
-  "-C", "linker=rust-lld",
   "-C", "link-arg=-Tlink.x",
 ]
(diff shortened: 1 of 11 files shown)


adamgreig:
Should we leave out `"-C", "linker=rust-lld"` entirely?

adamgreen:
I take it that **rust-lld** is the default linker anyway so this option is no longer required? If you guys decide that you would like to have that line removed from all of the .cargo/config files then I can make that change too. Just let me know.

adamgreen:
@adamgreig I have added commit 419ce796a121d0863721b19abca30f1b4ede02d2 to this PR to remove the ```"-C", "linker=rust-lld"``` flag completely from the .cargo/config files and mark down documentation.

I made sure that ```cargo build --target thumbv7em-none-eabihf``` at the root of the project still runs successfully and then took one of the resulting binaries and ran it on my Discovery board.

Thanks for the feedback!

adamgreig:
Great, thanks!

bors r+

PR #133, opened 2018-10-22
Switch '_ef' reference to 'ef' in GDB session example

adamgreen:
I see 'ef' used correctly everywhere else in this section. Looks like
this one just got missed in 2a213e70ebb763388cf344d3c678fe42dc7113d8.


adamgreig:
Urgh, this is the same mdbook 0.2.2 vs 0.2.1 issue as we've just had on bookshelf and book.

adamgreig:
bors retry

PR #138, opened 2018-10-28
A few documentation changes

adamgreen:
* I don't see the IOW acronym enough to know what it means without
  looking it up. I thought others might have the same issue so I
  expanded it. In the same sentence I changed what I think was an
  incorrect spelling of "whereas".
* I changed the text describing the itmdump of accelerometer readings
  to indicate that the actual output seen by the user if they run the
  same code won't be exactly as seen here.


thejpster:
Thanks :)

Issue #129, opened 2018-10-16
Add link to bookshelf and FAQs to front page


adamgreig:
We should add links back to the main bookshelf (docs.rust-embedded.org) and to the [FAQ](https://rust-embedded.github.io/bookshelf/faq.html) to the front page of the embedonomicon.

adamgreen:
I have attempted a fix in PR #140

adamgreig:
Closed by #140, thanks @adamgreen!

PR #140, opened 2018-10-30
Link back to bookshelf for other embedded resources

adamgreen:
This is my initial attempt to address issue #129 opened by @adamgreig
to have links back to the bookshelf added to the front page. I am
not sure if this is what you were looking for but I thought I would
give it a try. Let me know what you think.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -56,6 +56,16 @@ report it on the [issue tracker].
 [this repository]: https://github.com/rust-embedded/discovery
 [issue tracker]: https://github.com/rust-embedded/discovery/issues
 
+## Other embedded Rust resources
+
+This Discovery book is just one of several embedded Rust resources provided by the
+[Embedded Working Group]. The full selection can be found at [The Embedded Rust Bookshelf]. This
+includes the list of [Frequently Asked Questions].
+
+[Embedded Working Group]: https://github.com/rust-embedded/wg
+[The Embedded Rust Bookshelf]: https://rust-embedded.github.io/bookshelf/
+[Frequently Asked Questions]: https://rust-embedded.github.io/bookshelf/faq.html
+
 ## Sponsored by
 
 <p align="center">


src/README.md
```diff
@@ -56,6 +56,16 @@ report it on the [issue tracker].
 [this repository]: https://github.com/rust-embedded/discovery
 [issue tracker]: https://github.com/rust-embedded/discovery/issues
 
+## Other embedded Rust resources
+
+This Discovery book is just one of several embedded Rust resources provided by the
+[embedded Working Group]. The full selection can be found at ["The embedded Rust Bookshelf"]. This
+includes the list of [Frequently Asked Questions].
+
+[embedded Working Group]: https://github.com/rust-embedded/wg
```
therealprof: I find the casing of the word embedded a bit weird. Any reason not to use uppercase here...

src/README.md
```diff
@@ -56,6 +56,16 @@ report it on the [issue tracker].
...
+["The embedded Rust Bookshelf"]: https://rust-embedded.github.io/bookshelf/
```
therealprof: ... and here?

src/README.md
therealprof: I didn't mean that embedded should in general be spelled with capital 'E' but in the context of these links it looks a bit weird that everything is capitalised *except* for "embedded".

adamgreen: The two uses that you have flagged are spelled with that lowercase spelling on that page:
> Welcome to an overview of the documentation provided by the embedded Working Group. 

...
> Many of these resources take the form of "books"; we collectively call these "The embedded Rust Bookshelf".

...
> Also known as "the book" by the embedded Rust community. The embedded Rust book will ...

therealprof: Any suggestions how to normalise this in a sensible way? We can also defer this, it just stuck out while reviewing the PR.

therealprof: Looks fantastic.

adamgreen: Cool! I have pushed up that update to the PR.

Thanks for reviewing!

PR #145, opened 2018-11-07
Update bookshelf URLs to new location

adamgreen:
Documentation was recently moved to docs.rust-embedded.org and the old
links on this page now result in 404 errors.


adamgreen:
Thanks for reviewing and merging.

PR #173, opened 2019-02-23
Minor Linux guide fixes

Disasm:


Disasm:
Regarding `openocd-git`: it does work with my Discovery board and `openocd` doesn't. Looks like it's a bit outdated.

Disasm:
> Weird thing about openocd. Did a new release have a regression? The openocd package used to work.

Hmm. Now I can't reproduce this even with the different versions of ST-LINK firmware. I probably had bad udev rules that were installed a year ago (official ones come with `MODE="660", GROUP="plugdev"`) and now I have good ones.
I'm going to revert that `openocd-git` commit in favor of reproducible setup.

Issue #167, opened 2019-02-07
Brew cask issue installing gcc-arm-embedded


kylestew:
So you don't waste time trying to follow the instructions in the book to no end. The maintainers at the caskroom repo pulled this cask citing that is should live in Brew. It's not there yet. Head over to the ticket for some alternative actions. Hopefully it makes its way over to Brew soon.

https://github.com/Homebrew/homebrew-cask/pull/56802

chrisabruce:
TL;DR workaround

```bash
brew tap PX4/homebrew-px4
brew update

#list available
brew search px4

#latest as of this post
brew install gcc-arm-none-eabi-80
```

therealprof:
Any reason not to follow the instructions at https://rust-embedded.github.io/book/intro/install/macos.html ?

therealprof:
For me it would make sense to amend the book with missing information (if any) and then just point to the book.

ests:
I guess it's also possible to just follow instructions for Linux "other distros" ([link](https://docs.rust-embedded.org/discovery/03-setup/linux.html#other-distros)), but of course just download package for Mac OS X 64-bit.

kylestew:
Seems like there are enough solutions here to make this a non-issue.

lead-free:
Just tried `brew install gcc-arm-embedded` and besides installing the toolchain brew started pulling Keil.. ctrl+c, went to ARM website to get it:(

PR #179, opened 2019-03-12
Update macOS setup instructions for ARM toolchain.

ests:
The `gcc-arm-embedded` was deleted from the cask (pease see #167) and the setup command won't work. Closest working solution seems to be using formulae from [ARMmbed/homebrew-formulae](https://github.com/armmbed/homebrew-formulae).

### DIFF
--- a/src/03-setup/macos.md
+++ b/src/03-setup/macos.md
@@ -5,13 +5,13 @@ All the tools can be install using [Homebrew]:
 [Homebrew]: http://brew.sh/
 
 ``` console
-$ brew cask install gcc-arm-embedded
+$ # Arm GCC toolchain
+$ brew tap ArmMbed/homebrew-formulae
+$ brew install arm-none-eabi-gcc
 
+$ # Minicom and OpenOCD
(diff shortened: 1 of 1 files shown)


therealprof:
Maybe it would make sense to finally replace the duplicated installation instructions by a link to the book?
https://rust-embedded.github.io/book/intro/install.html

No need to maintain everything twice, no?

CC @rust-embedded/resources

src/03-setup/macos.md
```diff
@@ -5,14 +5,14 @@ All the tools can be install using [Homebrew]:
 [Homebrew]: http://brew.sh/
 
 ``` console
+$ # Arm GCC toolchain
+$ brew tap ArmMbed/homebrew-formulae
```
dmontagu: A plain
```
brew install armmbed/formulae/arm-none-eabi-gcc
```
worked for me, might be preferable to not add additional taps.

jamesmunns:
Merging manually, because bors is sad.

PR #186, opened 2019-06-12
Update verify.md

### LINKED ISSUE #183: Typo in 03-setup/verify.md: wrong permissions
Line 54 says:
```
As before, the permissions should be `crw-rw-r--`.
```
However, it should be:
```
As before, the permissions should be `crw-rw-rw-`.
```

derekdreery:
Closes #183


derekdreery:
Must be a spurious failure, I would think.

jamesmunns:
bors retry

PR #188, opened 2019-08-16
LED Roulette: Update 'flash it' docs

modulitos:
I encountered some confusion while working through section 5.2 "Flash It", so I've added some updates that future readers might find useful 🙏


thejpster:
Thanks!

PR #189, opened 2019-08-31
Update docs

modulitos:


src/06-hello-world/README.md
```diff
@@ -104,7 +104,7 @@ Breakpoint 1, main () at src/06-hello-world/src/main.rs:10
 10          let mut itm = aux6::init();
 ```
 
-Note that there's a `.gdbinit` at the root of the Cargo project. It's pretty similar to the one we
+Note that there's a `openocd.gdb` at the root of the Cargo project. It's pretty similar to the one we
```
modulitos: It looks like `openocd.gdb` used to be named `.gdbinit`, and this doc wasn't updated when the file name changed.

src/13-serial-over-bluetooth/README.md
```diff
@@ -22,4 +22,6 @@ Recommended steps to wire this up:
 And that's it! You should be able to run all the programs you wrote in [section 11] without
 modification! Just make sure you open the right serial device / COM port.
 
+**NOTE** If you are having trouble communicating with the bluetooth device, you may need to initialize USART1 with a lower baud rate. Lowering it from 115,200 bps to 9,600 bps might help, as described [here](https://github.com/rust-embedded/discovery/blob/master/src/11-usart/auxiliary/src/lib.rs#L31)
+
```
modulitos: I was stuck on this step, and after some debugging, I discovered that the program was getting stuck on this line:

```rust
while usart1.isr.read().rxne().bit_is_clear() {}
```

After some exploration - and some help from a friend! - I discovered that it was due to the baud mismatch between the chip and the bluetooth module.  Setting the baud rate to `9600.bps()`, as mentioned above, fixed the issue.

src/05-led-roulette/flash-it.md
```diff
@@ -102,7 +102,9 @@ Remote debugging using :3333
 0x00000000 in ?? ()
 ```
 
-**NOTE**: If you are getting errors like `undefined debug reason 7 - target needs reset` on the OpenOCD GDB server, then you may need to try using `arm-none-eabi-gdb` instead of the `gdb` command, as described above.
+**NOTE**: If you are getting errors like `undefined debug reason 7 - target needs reset`, you can try running `monitor reset halt` as described [here](https://stackoverflow.com/questions/38994596/reason-7-target-needs-reset-unreliable-debugging-setup).
+
+**NOTE**: If the debugger is still not connecting to the OpenOCD server, then you may need to try using `arm-none-eabi-gdb` instead of the `gdb` command, as described above.
```
modulitos: These changes are in response to this comment: https://github.com/rust-embedded/discovery/commit/ffeabc63d758085734bb7983ac3848b973c5f2ea#r34883585

My original comment incorrectly conflated the `undefined debug reason 7` output with my issue where I was using `gdb` instead of `arm-none-eabi-gdb`.  This update clarifies the issues that may arise, and addresses them separately.

PR #190, opened 2019-09-06
Updated install instructions for OSX

hickey:
The instructions for installing the ARM toolchain for OSX don't work any longer. The `gcc-arm-embedded` cask is not available in Homebrew any more. The enclosed changes update the documentation so others can install the toolchain on OSX.

### DIFF
--- a/src/03-setup/macos.md
+++ b/src/03-setup/macos.md
@@ -5,13 +5,63 @@ All the tools can be install using [Homebrew]:
 [Homebrew]: http://brew.sh/
 
 ``` console
-$ brew cask install gcc-arm-embedded
-
 $ brew install minicom openocd
 ```
 
-If the `brew cask` command doesn't work (`Error: Unknown command: cask`), then run `brew tap
-Caskroom/tap` first and try again.
+The `gcc-arm-embedded` are no longer available through [Homebrew]. The
+following is a step by step process for getting the ARM GCC cross
+compiler installed on OSX. Also available in Joe Goggins' [Gist] as
+referenced in [issue #118].
+
+[Gist]: https://gist.github.com/joegoggins/7763637
+[issue #118]: https://github.com/rust-embedded/book/issues/118
+
+Download the latest GNU Embedded Toolchain from the ARM Developer site
+using the following URL.
+
+    https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm/downloads
+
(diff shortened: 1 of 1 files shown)


hickey:
Interesting.... Another person here at the Seattle Rust meetup is using https://rust-embedded.github.io/book/intro/install/macos.html. Why is that location have different content from this repo?

hickey:
The page I was using is https://docs.rust-embedded.org/discovery/03-setup/macos.html

hickey:
The page I was using is https://docs.rust-embedded.org/discovery/03-setup/macos.html

hickey:
The page I was using is https://docs.rust-embedded.org/discovery/03-setup/macos.html

adamgreig:
Do the instructions from the Rust book work? We could just put those in if so, they seem simpler.

therealprof:
@hickey Any update on the question above?

jamesmunns:
Hey @hickey, I actually just accepted an update from #179, do those instructions still work? if not, I'd be happy to update to add your information, or make it a separate "build from source" option.

(I do think we need to de-dupe the setup instructions, but I'm happy to improve things for now).

hickey:
It has been so long since I have worked on any of the Rust stuff, I will have to rebuild the environment and test. I will try to do that soonish (if I am lucky it will be early next week). I will report back with what I have found.

src/03-setup/macos.md
```diff
@@ -5,13 +5,63 @@ All the tools can be install using [Homebrew]:
...
 
 ``` console
-
 $ brew install minicom openocd
 ```
 
-If the `brew cask` command doesn't work (`Error: Unknown command: cask`), then run `brew tap
-Caskroom/tap` first and try again.
+The `gcc-arm-embedded` are no longer available through [Homebrew]. The
```
ryankurte, line 11: ARM have an https://github.com/ARMmbed/homebrew-formulae which you can `brew tap ArmMbed/homebrew-formulae && brew install arm-none-eabi-gcc` that I tend to use for embedded dev and may be easier than this?

yerke:
There is no need for this PR any more, since `gcc-arm-embedded` cask is back (https://github.com/Homebrew/homebrew-cask/pull/75091).
On a related note, @jamesmunns , I see that you merged https://github.com/rust-embedded/discovery/pull/179, but I don't see the changes on https://docs.rust-embedded.org/discovery/03-setup/macos.html . I wonder why that would be case. Could it be that the book wasn't published/built to the site ever since?

PR #208, opened 2020-02-11
Fixed I2C registers page number of reference manual in Ch 14.3

mnauf:
Currently documentation refers to wrong page number here in Ch 14.3 which states that "Section 28.7 I2C registers - Page 873 - Reference Manual". I fixed it with page number 868.

### DIFF
--- a/src/14-i2c/read-a-single-register.md
+++ b/src/14-i2c/read-a-single-register.md
@@ -12,7 +12,7 @@ Just like with the USART peripheral, I've taken care of initializing everything 
 
 These registers are documented in the following section of the Reference Manual:
 
-> Section 28.7 I2C registers - Page 873 - Reference Manual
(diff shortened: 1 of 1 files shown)


thejpster:
Do you have a link to the reference manual? And actually it might be worth putting the link in the source.

Apologies in advance if it is there already - I couldn't see it on my phone after a brief look.

mnauf:
> Do you have a link to the reference manual? And actually it might be worth putting the link in the source.
> 
> Apologies in advance if it is there already - I couldn't see it on my phone after a brief look.
Sure here it is: http://www.st.com/resource/en/reference_manual/dm00043574.pdf
Its link is given in Chapter 7.1. Find it below
https://docs.rust-embedded.org/discovery/07-registers/rtrm.html

PR #212, opened 2020-02-18
Feat: add custom formatting for images

Rosto75:
Some images with alpha channel, are unreadable when a dark theme is applied. This commit adds a custom image style, named `white_bg`, that can be used to fix this issue.

The style is very simple, and adds two features:
- white background,
- a small padding around the image

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -3,3 +3,6 @@ title = "Discovery"
 description = "Discover the world of microcontrollers through Rust"
 author = "Rust Embedded Resources Team"
 language = "en"
+
+[output.html]
+additional-css = ["custom.css"]
--- a/book.toml
+++ b/book.toml
@@ -5,4 +5,4 @@ author = "Rust Embedded Resources Team"
 language = "en"
 
 [output.html]
-additional-css = ["./custom.css"]
+additional-css = ["custom.css"]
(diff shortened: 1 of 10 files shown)


adamgreig:
@Rosto75 looks like it's trying to find `custom.css` in every subdirectory, perhaps the entry in the configuration needs to be an absolute path?

Rosto75:
> @Rosto75 looks like it's trying to find `custom.css` in every subdirectory, perhaps the entry in the configuration needs to be an absolute path?

Any idea why the first check was successful, then? Can I replicate this failed test locally before I push a fix?

Rosto75:
Sorry for the extra commits. I was trying to squash them, unsuccessfully.

Rosto75:
I made some tests locally, and I believe that the issue is caused by the old version of mdBook used for CI check. The test script uses mdBook 0.2.x to generate book files, and links to the `custom.css` file, created by this version are broken. This is why the Linkchecker fails right after.

I suggest switching to the latest version of mdBook (0.3.5) and removing f7e6553, because the original PR is fine.

Rosto75:
I messed up my commit history, so I'll better start over, create a new PR, and run all integration tests before pushing it.

PR #213, opened 2020-02-24
Feat: add custom formatting for images

tomasz-rozanski:
Some images with alpha channel, are unreadable when a dark theme is applied. This commit adds a custom image style, named `white_bg`, that can be used to fix this issue.

The style is very simple, and adds two features:
- white background,
- a small padding around the image

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -3,3 +3,6 @@ title = "Discovery"
 description = "Discover the world of microcontrollers through Rust"
 author = "Rust Embedded Resources Team"
 language = "en"
+
+[output.html]
+additional-css = ["custom.css"]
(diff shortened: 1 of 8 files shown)


Rosto75:
bors check will fail because it uses `mdBook v0.2.1`, which generate invalid links to the custom css styles-sheets. This in turn causes `Linkchecker` to fail, and the whole build is unsuccessful.

The solution is to change the `ci/install.sh` script to use more recent version of `mdBook`.

adamgreig:
> The solution is to change the ci/install.sh script to use more recent version of mdBook.

Yep, on it.

tomasz-rozanski:
ping @rust-emdedded/resources for approval/merge

PR #225, opened 2020-04-27
Add knowledge requirement for hex and bitwise operations.

markhildreth:
Was wondering if this small addition to the "knowledge" requirements would be useful to add. I think having a full-on tutorial (even something in the appendix) for these would be a bit out of scope for this tutorial, but sometimes it helps to just give something a name for people to google later.

I'm not familiar with how technical book layouts like this works; is there some way that the code sample in this PR needs to be put into another file (e.g., so that there can be an automated check that the code still compiles), or would that only need to be reserved for larger, more important code samples?

### DIFF
--- a/src/02-requirements/README.md
+++ b/src/02-requirements/README.md
@@ -1,12 +1,29 @@
 # Hardware/knowledge requirements
 
-The only knowledge requirement to read this book is to know *some* Rust. It's
+The primary knowledge requirement to read this book is to know *some* Rust. It's
 hard for me to quantify *some* but at least I can tell you that you don't need
 to fully grok generics but you do need to know how to *use* closures. You also
 need to be familiar with the idioms of the [2018 edition], in particular with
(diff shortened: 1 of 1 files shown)


markhildreth:
@jamesmunns Pertaining to our [conversation on reddit](https://www.reddit.com/r/rust/comments/g7zzcm/rust_discovery_book_extremely_hard_to_follow/fomo1c0/), I started looking at some ways to improve the document, specifically around when the user starts to read the reference manuals. I'm starting small with this one. Not familiar with the general structure or best practices of mdbook projects, so feel free to get picking at all of the nits as I go :)

jamesmunns:
Hey @markhildreth, thanks for putting this together! I'll be doing some more review/triage later this week, feel free to ping me again if this sits unreviewed for more than a couple days :)

jamesmunns:
bors ping

(sorry for the noise)

jamesmunns:
bors r=me

PR #230, opened 2020-05-03
Removed overwhelming info, added context to "Meet your Hardware"

### LINKED ISSUE #199: RTRM question
On this page, https://docs.rust-embedded.org/discovery/07-registers/rtrm.html is this statement:

We are interested in the register that's at an offset of 0x18 from the base address of the GPIOE peripheral. According to the table, that would be the register BSRR.

Why are we interested in this offset?  How does one conclude that this is the correct offset / register for our uses.  I've RTRM's a lot and cannot find an obvious connection.

markhildreth:
In the first in a series of changes I'm proposing to try to help resolve issue #199, this changes the book's "Meet Your Hardware" in a few ways.

## Clearly delineate the board from the microcontroller.

While technically the original bullet points listed that the microcontroller was one of the components of the board, this hopefully makes a bit more clear that really they are two completely separate entities. This should especially help in Chapter 7 when readers are asked to jump between different reference manuals; understanding when to look at the reference manual of the board vs the microcontroller requires a solid understanding of the difference between these two products.

Also, given the book's expectation that the user has no previous knowledge of embedded development, I felt that listing specs (like RAM and Flash) probably only end up confusing them, as there is probably limited understanding of why they would need to know this information. Why is it important that the user knows that the microcontroller operates at 3.3v we're going to just plug in a USB cable and let the on-board voltage regulator do its job?

## Describe how "Arm" comes into all of this.

The book assumes that users have no familiarity with embedded development. I had _some_ experience (from the Arduino world) before working on a Feather M0, but still didn't understand what an "Arm processor" was, or why I was running tooling specifically made for Cortex-M chips. Hopefully this text (made as an aside so as to be easily skippable by those who do understand this delineation) can be useful.

### DIFF
--- a/src/04-meet-your-hardware/README.md
+++ b/src/04-meet-your-hardware/README.md
@@ -11,6 +11,7 @@ Let's get familiar with the hardware we'll be working with.
 We'll refer to this board as "F3" throughout this book. Here are some of the
 many components on the board:
 
+- A [microcontroller].
 - A number of LEDs, including the eight aligned in a "compass" formation.
 - Two buttons.
 - Two USB ports.
@@ -18,17 +19,17 @@ many components on the board:
 - A [magnetometer].
 - A [gyroscope].
 
+[microcontroller]: https://en.wikipedia.org/wiki/Microcontroller
 [accelerometer]: https://en.wikipedia.org/wiki/Accelerometer
 [magnetometer]: https://en.wikipedia.org/wiki/Magnetometer
 [gyroscope]: https://en.wikipedia.org/wiki/Gyroscope
 
-Of these components, the most important is the [microcontroller] (sometimes
+Of these components, the most important is the microcontroller (sometimes
 shortened to "MCU" for "microcontroller unit"), which is the large black square
 sitting in the center of your board. The MCU is what runs your code. You might
 sometimes read about "programming a board", when in reality what we are doing
 is programming the MCU that is installed on the board.
 
-[microcontroller]: https://en.wikipedia.org/wiki/Microcontroller
 
 ## STM32F303VCT6 (the "STM32F3")
 
--- a/src/04-meet-your-hardware/README.md
+++ b/src/04-meet-your-hardware/README.md
@@ -42,17 +42,20 @@ current flows through a circuit. By enabling or disabling electrical current to
 flow through a specific pin, an LED attached to that pin (via the traces) can
 be turned on and off.
 
-Part numbers of components can often give us information about the part. In the
-case of our MCU (with part number `STM32F303VCT6`), the `ST` at the start
-hints to us that this is manufactured by [ST Microelectronics]. Each
-manufacturer uses a different part numbering scheme, but with some research we
-can also determine that the `M32` implies that this is specifically an
-Arm®-based 32-bit microcontroller, and the `F3` indicates that the
-microcontroller is based on the Cortex®-M4. The rest of the part number
-provides more specifics (e.g., extra features, RAM size), and aren't as
-important to us right now.
+Each manufacturer uses a different part numbering scheme, but many will allow
+you to determine information about a component simply by looking at the part
+number. Looking at our MCU's part number (`STM32F303VCT6`), the `ST` at the
+front hints to us that this is a part manufactured by [ST Microelectronics].
+Searching through [ST's marketing materials] we can also learn the following:
 
 [ST Microelectronics]: https://st.com/
+[ST's marketing materials]: https://www.st.com/en/microcontrollers-microprocessors/stm32-mainstream-mcus.html
+
+- The `M32` represents that this is an Arm®-based 32-bit microcontroller.
+- The `F3` represents that the MCU is from ST's "STM32F3" series. This is a
+  series of MCUs based on the Cortex®-M4 processor design.
+- The remainder of the part number goes into more details about things like
+  extra features and RAM size, which at this point we're less concerned about.
 
 > ### Arm? Cortex-M4?
 >
--- a/src/04-meet-your-hardware/README.md
+++ b/src/04-meet-your-hardware/README.md
@@ -43,10 +43,10 @@ flow through a specific pin, an LED attached to that pin (via the traces) can
 be turned on and off.
 
 Part numbers of components can often give us information about the part. In the
-case of our MCU (with part number `STM32F303VCT6`), the `STM` at the start
+case of our MCU (with part number `STM32F303VCT6`), the `ST` at the start
 hints to us that this is manufactured by [ST Microelectronics]. Each
 manufacturer uses a different part numbering scheme, but with some research we
-can also determine that the `32` implies that this is specifically an
+can also determine that the `M32` implies that this is specifically an
 Arm®-based 32-bit microcontroller, and the `F3` indicates that the
 microcontroller is based on the Cortex®-M4. The rest of the part number
 provides more specifics (e.g., extra features, RAM size), and aren't as
--- a/src/04-meet-your-hardware/README.md
+++ b/src/04-meet-your-hardware/README.md
@@ -30,7 +30,6 @@ sitting in the center of your board. The MCU is what runs your code. You might
 sometimes read about "programming a board", when in reality what we are doing
 is programming the MCU that is installed on the board.
 
-
 ## STM32F303VCT6 (the "STM32F3")
 
 Since the MCU is so important, let's take a closer look at the one sitting on our board.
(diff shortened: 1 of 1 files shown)


src/04-meet-your-hardware/README.md
```diff
@@ -8,40 +8,79 @@ Let's get familiar with the hardware we'll be working with.
...
+is programming the MCU that is installed on the board.
+
+[microcontroller]: https://en.wikipedia.org/wiki/Microcontroller
+
+## STM32F303VCT6 (the "STM32F3")
+
+Since the MCU is so important, let's take a closer look at the one sitting on our board.
+
+Our MCU is surrounded by 100 tiny metal **pins**. These pins are connected to
+**tracings**, the little "roads" that act as the wires connecting components
```
adamgreig, line 38:
```suggestion
**traces**, the little "roads" that act as the wires connecting components
```

src/04-meet-your-hardware/README.md
```diff
@@ -8,40 +8,79 @@ Let's get familiar with the hardware we'll be working with.
...
-- An [accelerometer] and a [magnetometer][] (in a single package).
+We'll refer to this board as "F3" throughout this book. Here are some of the
+many components on the board:
+
+- A number of LEDs, including the eight aligned in a "compass" formation.
+- Two buttons.
+- Two USB ports.
+- An [accelerometer].
+- A [magnetometer].
+- A [gyroscope].
```
adamgreig, line 19: I appreciate the list isn't meant to be exhaustive but since you later say "of _these_ components, the most important is the microcontroller", perhaps the microcontroller should be mentioned in this list, perhaps even at the top.

src/04-meet-your-hardware/README.md
```diff
@@ -8,40 +8,79 @@ Let's get familiar with the hardware we'll be working with.
...
+flow through a specific pin, an LED attached to that pin (via the tracing) can
+be turned on and off.
+
+By looking at the part number and doing some research, we can learn more
+information about electronic components. The part number of our microcontroller
+starts with `STM32`, which indicates that it was manufactured by [ST
+Microelectronics]. It also lets us know that this is an Arm®-based 32-bit
+microcontroller. The `F3` in the part number indicates that the microcontroller
+is based on the Cortex®-M4. The rest of the part number provides more specifics
+(e.g., extra features, RAM size), and aren't as important to us right now.
```
adamgreig, line 51: Do you think it's worth linking to the [product page](https://www.st.com/en/microcontrollers-microprocessors/stm32f303vc.html) for this MCU?

src/04-meet-your-hardware/README.md
```diff
@@ -8,40 +8,79 @@ Let's get familiar with the hardware we'll be working with.
...
+## STM32F303VCT6 (the "STM32F3")
+
+Since the MCU is so important, let's take a closer look at the one sitting on our board.
+
+Our MCU is surrounded by 100 tiny metal **pins**. These pins are connected to
+**tracings**, the little "roads" that act as the wires connecting components
+together on the board. The MCU can dynamically alter the electrical properties
+of the pins. This works similar to a lightswitch altering how electrical
+current flows through a circuit. By enabling or disabling electrical current to
+flow through a specific pin, an LED attached to that pin (via the tracing) can
```
jamesmunns, line 42:
```suggestion
flow through a specific pin, an LED attached to that pin (via the traces) can
```

src/04-meet-your-hardware/README.md
jamesmunns: We may also want to note that this naming scheme (and the information you pulled from it), varies between manufacturers, which each have their own naming conventions.

src/04-meet-your-hardware/README.md
```diff
@@ -8,40 +8,79 @@ Let's get familiar with the hardware we'll be working with.
...
+[ST Microelectronics]: https://st.com/
+
+> ### Arm? Cortex-M4?
+>
+> If our chip is manufactured by ST, then who is Arm? And if our chip is the
+> STM32F3, what is the Cortex-M4?
+>
+> You might be surprised to hear that while "Arm-based" chips are quite
+> popular, the company behind the "Arm" trademark ([Arm Holdings][]) doesn't
+> actually manuafacture chips for purchase. Instead, their primary business
```
jamesmunns, line 62:
```suggestion
> actually manufacture chips for purchase. Instead, their primary business
```

src/04-meet-your-hardware/README.md
markhildreth: Good point. Added in a22fa6017d9437c46f51172f161761e802d6af45.

src/04-meet-your-hardware/README.md
markhildreth: @jamesmunns That seems like good info to have. I tried to tweak the language to include it in https://github.com/rust-embedded/discovery/pull/230/commits/042dfbb28bc4b970d5448af6ef6f5299ff363ee1.

@adamgreig I wouldn't be against it. My only concern is that it's a URL to the site probably controlled by their marketing department, which (in my experience, at least) is more likely to  be left broken at some point than, say, the URL to a datasheet behind a "/resources" path.

src/04-meet-your-hardware/README.md
```diff
@@ -43,13 +43,15 @@ current flows through a circuit. By enabling or disabling electrical current to
...
-(e.g., extra features, RAM size), and aren't as important to us right now.
+Part numbers of components can often give us information about the part. In the
+case of our MCU (with part number `STM32F303VCT6`), the `STM` at the start
+hints to use that this is manufactured by [ST Microelectronics]. Each
+manufacturer uses a different part numbering scheme, but with some research we
+can also determine that the `32` implies that this is specifically an
+Arm®-based 32-bit microcontroller, and the `F3` indicates that the
+microcontroller is based on the Cortex®-M4. The rest of the part number
+provides more specifics (e.g., extra features, RAM size), and aren't as
+important to us right now.
```
adamgreig, line 53: I think it's more accurate to say the `ST` at the start tells us it's made by ST, while the `M32` tells us it's a 32-bit microcontroller, and the `F3` tells us it's in ST's `STM32F3` family of Cortex-M4-based microcontrollers. I concede your point on the link; ST have definitely changed them in the past and it's pretty easily found with a search anyway.

markhildreth: Ah, thanks. Changed in 018702e0332ab864e7e4a59b4a2f49bbd3d78669. Let me know if that makes more sense.

adamgreig: Thanks! My only remaining quibble is that `F3` by itself doesn't tell us that it's based on a Cortex-M4 core (you might even expect it to be a Cortex-M3 core given the `3`); it tells us it's part of the STM32F3 family and you just have to know that the F3 family uses a Cortex-M4 core.

markhildreth: Ah, okay, I see what you're saying. Yeah, I can see how that could be confusing, and why the link would be helpful.

How about something like this:

> Each manufacturer uses a different part numbering scheme. Looking at our MCU's part number (`STM32F303VCT6`), the `ST` at the front hints to us that this is a part manufactured by [ST Microelectronics](https://www.st.com/). Searching through [ST's marketing materials](https://www.st.com/en/microcontrollers-microprocessors/stm32-mainstream-mcus.html), we can also learn the following:
> * The `M32` represents that this is an Arm®-based 32-bit microcontroller.
> * The `F3` represents that the microcontroller is based on the Cortex®-M4 processor design.
> * The remainder of the part number goes into more details about things like extra features and RAM size, which at this point we're less concerned about.

markhildreth: Sorry, meant to also say: yes, the link has the chance to break. But as you mentioned, it's easy to search for even if the link breaks.

adamgreig: The second link does look better. I guess I'm not sure about saying the `F3` "represents" that it's based on a Cortex-M4, since you can't infer that from the `F3`. Perhaps something like...

> * The M32 means this is one of ST's Arm®-based STM32 32-bit microcontrollers.
> * The F3 means this microcontroller is in ST's STM32F3 series, which use Arm's Cortex®-M4 processor design.

markhildreth: Gotcha, yeah I seem to keep skipping that step, huh? Thanks for keeping me honest. I've tried to put this together in 524530728410f10b1231d51b94d9880f0faec115, this time meantioning that F3 describes a series of MCU, and that the series utilizes the Cortex-M4.

markhildreth:
@adamgreig No worries, it definitely helped. Thanks for the fast feedback!

Issue #10, opened 2016-11-13
GDB cheatsheet


japaric:

jamesmunns:
This might be a reasonable start: https://github.com/ferrous-systems/embedded-trainings/blob/master/beginner/docs/segment-1.md#part-4-debugging-code

yusefkarim:
I have added a section to the appendix.

I am not an expert with GDB nor am I a concise writer. If you like these changes and see errors or things you would like added, please let me know!

eldruin:
Added in #234

PR #234, opened 2020-05-04
Gdb usage appendex

yusefkarim:
Refers to #10

### DIFF
--- a/src/appendix/2-how-to-use-gdb/README.md
+++ b/src/appendix/2-how-to-use-gdb/README.md
@@ -4,33 +4,35 @@ Below are some useful GDB commands that can help us debug our programs. This ass
 
 ## General Debugging
 
-> **NOTE:** Many of the commands you see below can be executed using a short form. For example, `continue` can simply be used as `c`, or `break $location` can be used as `b $location`. While you are learning GDB try to see how short you can get the commands to go before GDB doesn't recognize it!
+> **NOTE:** Many of the commands you see below can be executed using a short form. For example, `continue` can simply be used as `c`, or `break $location` can be used as `b $location`. Once you have experience with the commands below, try to see how short you can get the commands to go before GDB doesn't recognize it!
 
 
 ### Dealing with Breakpoints
 
 * `break $location`: Set a breakpoint at a place in your code. The value of `$location` can include:
-    * `b 123` - Break on line 123 of the currently displayed file
-    * `b main.rs:123` - Break on line 123 of `main.rs`
+    * `break *main` - Break on the exact address of the function `main`
+    * `break *0x080012f2` - Break on the exact memory location `0x080012f2`
+    * `break 123` - Break on line 123 of the currently displayed file
+    * `break main.rs:123` - Break on line 123 of the file `main.rs`
 * `info break`: Display current breakpoints
 * `delete`: Delete all breakpoints
-    * `d $n`: Delete breakpoint `$n`
+    * `delete $n`: Delete breakpoint `$n`
 * `clear`: Delete breakpoint at next instruction
     * `clear main.rs:$function`: Delete breakpoint at entry of `$function` in `main.rs`
     * `clear main.rs:123`: Delete breakpoint on line 123 of `main.rs`
 * `enable`: Enable all set breakpoints
-  * `en $n`: Enable breakpoint `$n`
+  * `enable $n`: Enable breakpoint `$n`
 * `disable`: Disable all set breakpoints
-  * `dis $n`: Disable breakpoint `$n`
+  * `disable $n`: Disable breakpoint `$n`
 
 ### Controlling Execution
 
-* `continue`: Begin execution of your program
+* `continue`: Begin or continue execution of your program
 * `next`: Execute the next line of your program
-    * `n $n`: Repeat `next` `$n` number times
+    * `next $n`: Repeat `next` `$n` number times
 * `nexti`: Same as `next` but with machine instructions instead
 * `step`: Execute the next line, if the next line includes a call to another function, step into that code
-    * `s $n`: Repeat `step` `$n` number times
+    * `step $n`: Repeat `step` `$n` number times
 * `stepi`: Same as `step` but with machine instructions instead
 * `jump $location`: Resume execution at specified location:
     * `jump 123`: Resume execution at line 123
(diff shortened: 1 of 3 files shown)


src/appendix/2-how-to-use-gdb/README.md
```diff
@@ -0,0 +1,87 @@
+# How to use GDB
+
+Below are some useful GDB commands that can help us debug our programs. This assumes you have [flashed a program](../../05-led-roulette/flash-it.md) onto your microcontroller and attached to an OpenOCD session.
+
+## General Debugging
+
+> **NOTE:** Many of the commands you see below can be executed using a short form. For example, `continue` can simply be used as `c`, or `break $location` can be used as `b $location`. While you are learning GDB try to see how short you can get the commands to go before GDB doesn't recognize it!
```
hargoniX, line 7: Not quite sure about this comment, users might end up with some confusing behaviour they don't really understand, I'd say getting the short forms of the command is not something for a person who's just learning GDB.

src/appendix/2-how-to-use-gdb/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+
+### Dealing with Breakpoints
+
+* `break $location`: Set a breakpoint at a place in your code. The value of `$location` can include:
+    * `b 123` - Break on line 123 of the currently displayed file
+    * `b main.rs:123` - Break on line 123 of `main.rs`
```
hargoniX, line 14: I'd say we should mention that it is possible to use break like `b function_name` or even at specific locations of memory etc, refer to the gdb help text.

hargoniX: Also if we can agree to remove the shortcuts from the material for now just to make it more clear, after all they aren't being used consistently right now either, e.g. `i b` is not mentioned along `info break`.

src/appendix/2-how-to-use-gdb/README.md
```diff
@@ -0,0 +1,87 @@
...
+* `info break`: Display current breakpoints
+* `delete`: Delete all breakpoints
+    * `d $n`: Delete breakpoint `$n`
```
hargoniX, line 17: Same not on shortcuts

src/appendix/2-how-to-use-gdb/README.md
```diff
@@ -0,0 +1,87 @@
...
+* `clear`: Delete breakpoint at next instruction
+    * `clear main.rs:$function`: Delete breakpoint at entry of `$function` in `main.rs`
+    * `clear main.rs:123`: Delete breakpoint on line 123 of `main.rs`
+* `enable`: Enable all set breakpoints
+  * `en $n`: Enable breakpoint `$n`
```
hargoniX, line 22: Shortcuts

src/appendix/2-how-to-use-gdb/README.md
```diff
@@ -0,0 +1,87 @@
...
+* `disable`: Disable all set breakpoints
+  * `dis $n`: Disable breakpoint `$n`
```
hargoniX, line 24: Shortcuts

src/appendix/2-how-to-use-gdb/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Controlling Execution
+
+* `continue`: Begin execution of your program
```
hargoniX, line 28: I'd prefer something like "Begin or continue execution of your program" but I guess that's obvious from the name...

src/appendix/2-how-to-use-gdb/README.md
```diff
@@ -0,0 +1,87 @@
...
+* `next`: Execute the next line of your program
+    * `n $n`: Repeat `next` `$n` number times
```
hargoniX, line 30: Shorcuts

src/appendix/2-how-to-use-gdb/README.md
```diff
@@ -0,0 +1,87 @@
...
+* `nexti`: Same as `next` but with machine instructions instead
+* `step`: Execute the next line, if the next line includes a call to another function, step into that code
+    * `s $n`: Repeat `step` `$n` number times
```
hargoniX, line 33: Shorcuts

src/appendix/2-how-to-use-gdb/README.md
```diff
@@ -0,0 +1,87 @@
...
+    * `x /5i 0x080012c4`: Print 5 machine instructions staring at address `0x080012c4`
+    * `x/4xb $pc`: Print 4 bytes of memory starting where `$pc` currently is pointing
+* `disassemble $location`
+    * `disas /r main`: Disassemble the function `main`, using `/r` to show the bytes that make up each instruction
+
+
+### Looking at the Symbol Table
+
+* `info functions $regex`: Print the names and data types of functions matched by `$regex`, omit `$regex` to print all functions
+    * `i func main`: Print names and types of defined functions that contain the word `main`
```
hargoniX, line 63: Shorthand...I'll stop mentioning them from now on.

hargoniX (changes requested):
Despite my lots of comments most of them are really just about the use of shortcuts which I don't feel are good for a cheatsheet for a beginner, they can figure out the shortcuts themselves if it is mentioned at the beginning that they exist and are usually mnemonics such as b(reak). Encouraging them to use shortcuts though just feels like a potential error and confusion source for me. Apart from that there is really only the comment on how to break, other than that great work and a good addition!

src/appendix/2-how-to-use-gdb/README.md
yusefkarim: You may be right, I have not removed it yet, but made a slight change in addition to getting rid of all the shortcuts in the main section. If you still think it could lead to any confusion, let me know and we can take it out. Thanks.

src/appendix/2-how-to-use-gdb/README.md
yusefkarim: I added two more examples, could likely squish these into two again but have source filename, function name, label name, and line number all explained (like in help text). These would be more concise but maybe a little harder to read. Any preference?

src/appendix/2-how-to-use-gdb/README.md
yusefkarim: Better to be explicit, I added your suggestion. Nice catch.

src/appendix/2-how-to-use-gdb/README.md
```diff
@@ -0,0 +1,89 @@
+# How to use GDB
+
+Below are some useful GDB commands that can help us debug our programs. This assumes you have [flashed a program](../../05-led-roulette/flash-it.md) onto your microcontroller and attached to an OpenOCD session.
+
+## General Debugging
+
+> **NOTE:** Many of the commands you see below can be executed using a short form. For example, `continue` can simply be used as `c`, or `break $location` can be used as `b $location`. Once you have experience with the commands below, try to see how short you can get the commands to go before GDB doesn't recognize it!
```
eldruin, line 7:
```suggestion
> **NOTE:** Many of the commands you see below can be executed using a short form. For example, `continue` can simply be used as `c`, or `break $location` can be used as `b $location`. Once you have experience with the commands below, try to see how short you can get the commands to go before GDB doesn't recognize them!
```

src/appendix/2-how-to-use-gdb/README.md
```diff
@@ -0,0 +1,89 @@
...
+### Dealing with Breakpoints
+
+* `break $location`: Set a breakpoint at a place in your code. The value of `$location` can include:
+    * `break *main` - Break on the exact address of the function `main`
+    * `break *0x080012f2` - Break on the exact memory location `0x080012f2`
+    * `break 123` - Break on line 123 of the currently displayed file
+    * `break main.rs:123` - Break on line 123 of the file `main.rs`
+* `info break`: Display current breakpoints
+* `delete`: Delete all breakpoints
+    * `delete $n`: Delete breakpoint `$n`
```
eldruin, line 19:
```suggestion
    * `delete $n`: Delete breakpoint `$n` (`n` being a number. For example: `delete $2`)
```

eldruin (changes requested):
Great addition! Just a couple nitpicks from my side.

eldruin (approved):

eldruin:
Fine for me now but let's wait for @hargoniX who was much more involved here.

hargoniX (approved):
LGTM to me as well!

eldruin:
Great, then let's get this merged!

bors r+

PR #238, opened 2020-05-14
+= line numbers in release build to allow stepping

Darneas:
[Section 7.2](https://docs.rust-embedded.org/discovery/07-registers/optimization.html) cannot be reproduced with the provided files as line numbers are missing in release build.

This change adds them by modifying the release profile.

### DIFF
--- a/src/07-registers/Cargo.toml
+++ b/src/07-registers/Cargo.toml
@@ -4,5 +4,8 @@ edition = "2018"
 name = "registers"
 version = "0.1.0"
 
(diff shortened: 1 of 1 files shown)


BartMassey:
Thanks much for this PR! Sorry it's taking so long to merge.

BartMassey (approved):
This is a good idea. Rebases trivially. Should be rebased and merged.

eldruin (approved):
Thank you!

PR #239, opened 2020-05-18
Added contextual sections on memory mapping and peripherals

markhildreth:
This PR covers steps 2 and 3 in [my plans to resolve issue #199](https://github.com/rust-embedded/discovery/issues/199#issuecomment-620783270). It does the following:

1.) Clean up the introductory page to section 7, making it much shorter and more to the point that we are not expecting the user to fully grok the code yet, but we will be explaining it soon. This includes moving the aside into a later section.

2.) Add in additional sections ("Memory Mapping" and "Peripherals") that go into more detail about what is going on under the hood. This hopefully explains to the user why accessing some random value from memory will have the effect that it does, and provide more context for what exactly it is they are looking to get out of the reference manuals.

My plan is if this PR gets merged, I will submit a final PR which tries to make the RTRM section a bit easier to understand as well, building off of the context provided in these two new sections.


markhildreth:
@adamgreig Any status on this? I understand it's text-dense, but just making sure it wasn't lost in the shuffle.

jamesmunns:
r? me

Tagging so I remember to check this, @adamgreig feel free to take it back if you get to it before me :)

src/07-registers/memory-mapping.md
```diff
@@ -0,0 +1,115 @@
...
+```
+0x0000_0000: Access the first byte (`0x00`) of Flash.
+0x0000_0003: Access the fourth byte (`0x03`) of Flash.
+0x0003_FFFF: Access the final byte of Flash.
+0x0004_0000: Access the first byte (`0x00`) of SRAM.
+0x0004_0004: Access the fourth byte (`0x04`) of SRAM.
+0x0004_BFFF: Access the final address of SRAM.
+```
+
+As a result, manufacturers can what types and sizes of memory to to build into
```
adamgreig, line 89:
```suggestion
As a result, manufacturers can decide what types and sizes of memory to to build into
```

src/07-registers/memory-mapping.md
```diff
@@ -0,0 +1,115 @@
...
+The solution used by the designers of our processor is to assign all of the
+addresses from both memory types onto one global address space. For example, we
+could create a simple mapping of values that looks like this:
+
+```
+Addresses 0x0000_0000 - 0x0003_FFFF: 256kb Flash
+Addresses 0x0004_0000 - 0x0004_BFFF: 48kb SRAM
+```
+
+> This isn't the actual map for your controller, but useful as an example.
```
adamgreig, line 75: Could we just use the real memory map from the F3? I don't think it's any more complicated, you can even use 0x0 for flash since it's accessible from address 0 in the default map.

markhildreth: It's probably a good idea to do this. However, there are a few things that make the real version more complex:

* The 48kb SRAM is actually split between 40kb SRAM at `0x2000 0000 - 0x2000 9FFF` and 8k CCRAM at `0x1000 0000 - 0x1000 1FFF`. For multiple reasons (including my inadequate knowledge in what the heck CCRAM is used for), I decided to try to simplify.
* I don't actually own one of these, and was confused when I saw this in that the Flash memory is listed twice in the reference manual memory map :)

image

markhildreth:
@jamesmunns @adamgreig Any further thoughts?

hargoniX:
Since we agreed that this book is not working on the right level in [this weeks meeting](https://github.com/rust-embedded/wg/blob/master/minutes/2020-08-25.md) we will refrain from merging this PR for now, however at some point the low level information will be moved on to the other books, it would be cool if you could add your changes there once we get to that point!

markhildreth:
@hargoniX Thanks for taking a look at this. When I first started looking into this, I tried to take an incremental approach, but the more I worked the more I came to the same realization that, long run, reorganizing the books might be best. Is there any area documenting this effort to reorganize the books?

hargoniX:
Presently such an issue does not exist no, if you however feel like more in the books than discovery book chapter 7+ should be reorganized I guess it would be fair to open an issue over in the wg repository to talk about that in general and maybe we can bring it up during the next meeting.

Issue #263, opened 2020-08-24
 STM32F3DISCOVERY replacement


bitdivine:
Hello all,

Given that the board that he book is based on is no longer available, is it worth making an updated version?  I imagine the easy answer is yes but the next question is what the replacement development board should be.  One from the NUCLEO series?  The STM32F3348-DISCO?

Another question is how to manage multiple copies of the book with different boards.  A branch per board?

Note: I am happy to do the work, but it would be good to have alignment before I start.

Best wishes.

hargoniX:
I was gonna look into rewriting large parts of this books as per #262. Without actually having one of these discovery boards it seems kinda hard to me to actually figure out whether my rewrite works though so a big :+1: from my side on using a widely available board being used for this. Until we switch or I get my hands on one I will be kind of blocked on this sadly.

theunkn0wn1:
If the discovery board is not obtainable, my suggestion would be to target a nucleo board.
I found the nucleo-64 with the stm32f446retx on it to be easily obtainable. The question then becomes finding a good handful of sensors and the sort to go with.

David-OConnor:
Do you have a source on it being NLA? I bought one off Amazon (US) a few days ago.

hargoniX:
I talked to another person on Discord yesterday who also reported not being able to find anything online. And just now i went through the first two pages or duck duck go results for the board and couldnt get anything in stock there either, not even at big companies like Mouser.

So even if it might be in stock somewhere it doesnt seem to me like it is trivial to acquire one, hence Id definitely be in for a change.

sebpuetz:
Just chiming in here: I tried to get one of these boards (located in Germany) to start the guide and I can't get my hands on any. It's either noted as out-of-stock or delivery date in 4+ weeks.

hargoniX:
It has been agreed during the wg meeting that we will be looking into using the micro:bit for the rewrite so stay tuned for that.

Closing as this resolves the original question asked in the issue

black-puppydog:
@sebpuetz for what it's worth, I was also put off by the 4 week delivery date. ordered it anyhow, thinking I probably wouldn't get to it sooner anyhow, and got it within a few days... :P

bitdivine:
Mouser and digi-key have these in stock again.  Perhaps just a supply chain disruption.

shifty331:
[...]

PR #267, opened 2020-09-04
fix for 11-usart

### LINKED ISSUE #264: 11-usart None on Peripherals take
Hi there,
I'm trying to run an example code from `11-usart` section without using external serial module;
Unfortunately, I still cannot get "x" printed in Minicom, also I've noticed that `11-usart/auxiliary/src/lib.rs:18` which is `let cp = cortex_m::Peripherals::take().unwrap();` panics with `None`.

Could you please help me with that?

zhao-lang:
@paulkernfeld should fix #264 

`11-usart` was panicking on the older version of `cortex-m`, updating to `0.6.3` fixes that, but introduces incompatibilities with the `f3` crate. I replaced the `f3` crate with the newer `stm32f3xx-hal`. Also had to add `memory.x` to project root in order to get things to compile, maybe there's a better solution for that?


eldruin:
FYI, we are looking at a major rewrite of this book since the STM32F3Discovery boards have become difficult to buy. We may choose the [micro:bit](https://tech.microbit.org/hardware/) board as a replacement. See #263

andre-richter:
I am not familiar with the cortex-m stuff. Temporarily switching my assignment mit @adamgreig. Adam, please see if you can review or delegate this. Thanks :)

bitdivine:
@andre-richter Is someone else available to review this?  I assume that @adamgreig is busy.

adamgreig:
Ugh, sorry, this totally fell off my radar. I think there's two issues merging this as-is:

1. It requires an additional change to `11-usart/src/main.rs` to put the call to `tdr().bits()` into an unsafe block (and also removing the `#[deny(unsafe_code)]` attribute at the top), but sticking that unsafe code in seems quite confusing to newcomers and would need explaining (why do we need it here but not elsewhere, etc)
2. I guess memory.x was being provided by the HAL, but we shouldn't then stick a new one in the repository root. We could probably place it inside the `11-usart/` directory along with a `build.rs` script to copy it to the linker directory.

For the second point I think we can just swap the dependency to the new `stm32f3-discovery` crate and update the exports to `stm32f3_discovery::stm32f3xx_hal::...`, so we don't need a `memory.x` at all.

However I think we still need the unsafe block which isn't great... looking into it it seems like an error in the stm32f3 crate where the TDR register is said to be safe for any value 0-255 but as it's a nine-bit register it should actually go to 0x1ff. Dunno why that's not been caught before to be honest... but it won't be a quick update, so for now I guess we're stuck with the unsafe block.

adamgreig:
This has now been fixed by #354, thank you for starting the work in this PR!

PR #275, opened 2020-10-31
06-hello-world: change note about solder bridge

julianoes:
I got slightly confused by the note saying that the datasheet was wrong
about the solder bridge because from how I read it, it says:

Screenshot_2020-10-31 Discovery kit with STM32F303VC MCU - User manual - dm00063382-discovery-kit-with-stm32f303vc-mcu-stmi

I wonder if this might have been fixed in the datasheet in the meantime?

### DIFF
--- a/src/06-hello-world/README.md
+++ b/src/06-hello-world/README.md
@@ -1,8 +1,8 @@
 # Hello, world!
 
-> **HEADS UP** Several readers have reported that the "solder bridge" SB10 (see back of the board)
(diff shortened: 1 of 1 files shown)


julianoes:
Ping :smiley:

adamgreig (approved):
Thanks, there's nothing in the revision history of the manual about this but I don't have old versions to compare to. In any event clearly the currently manual does say it should be off by default.

bors merge

julianoes:
Thanks @adamgreig!

PR #277, opened 2020-11-15
Rewrite serial

hargoniX:
This PR:

* gets rid off the old chapters explaining how registers etc. work since a person starting out with embedded rust doesn't have to know that anymore
* moves the UART and serial communication chapters a little forward since they seemed like the most simple of hardware peripherals to me (after LEDs that is).
* changes the UART code from register level code to nrf51-hal code

Everything is rewritten now.


src/06-serial-communication/README.md
```diff
@@ -21,9 +21,11 @@ signal. Instead both parties must agree on how fast data will be sent along the
...
 and from B to A simultaneously.
 
-We'll be using this protocol to exchange data between the microcontroller and your computer. In
-contrast to the ITM protocol we have used before, with the serial communication protocol you can
-send data from your computer to the microcontroller.
+We'll be using this protocol to exchange data between the microcontroller and your computer. Now you might
+asking yourself why exactly we aren't using RTT for this like we did before. RTT is a protocol that is meant
+to be used soley for debugging. You will most definitely not be able to find a device that actually uses RTT
+to communicate with some other device in production. However serial communication is used quite often, for
+example some GPS receivers send the positioning information they receive to your via serial communication.
```
eldruin, line 28:
```suggestion
example some GPS receivers send the positioning information they receive via serial communication.
```

src/06-serial-communication/README.md
```diff
@@ -39,10 +41,10 @@ rate will probably be lower because of processing times on the slower side of th
 microcontroller).
 
 Today's computers don't support the serial communication protocol. So you can't directly connect
-your computer to the microcontroller. But that's where the serial module comes in. This module will
-sit between the two and expose a serial interface to the microcontroller and an USB interface to
-your computer. The microcontroller will see your computer as another serial device and your computer
-will see the microcontroller as a virtual serial device.
+your computer to the microcontroller. Luckily for us thought the debug probe on the micro:bit has a so called
```
eldruin, line 44:
```suggestion
your computer to the microcontroller. Luckily for us though, the debug probe on the micro:bit has a so-called
```

src/06-serial-communication/README.md
```diff
@@ -39,10 +41,10 @@ rate will probably be lower because of processing times on the slower side of th
...
+USB to serial converter. What this means is that it will sit between the two and expose a serial interface to
```
eldruin, line 45:
```suggestion
USB to serial converter. What this means is that the converter will sit between the two and expose a serial interface to
```

src/06-serial-communication/nix-tooling.md
```diff
@@ -19,42 +19,9 @@ the board; check the next section, which contains instructions for older
...
 "minicom" section.
 
-## Older revisions of the discovery board / external serial module
+## minicom
 
-Connect the serial module to your computer and let's find out what name the OS assigned to it.
-
-> **NOTE** On macs, the USB device will named like this: `/dev/cu.usbserial-*`. You won't
-> find it using `dmesg`, instead use `ls -l /dev | grep cu.usb` and adjust the following 
-> commands accordingly!
```
eldruin, line 28: This is still useful for macOS

src/06-serial-communication/nix-tooling.md
```diff
@@ -19,42 +19,9 @@ the board; check the next section, which contains instructions for older
...
-
-But what's this `ttyUSB0` thing? It's a file of course! Everything is a file in \*nix:
-
-``` console
-$ ls -l /dev/ttyUSB0
-crw-rw-rw- 1 root uucp 188, 0 Oct 27 00:00 /dev/ttyUSB0
-```
-
-> **NOTE** if the permissions above is `crw-rw----`, the udev rules have not been set correctly
-> see [udev rules](../03-setup/linux.html#udev-rules)
```
eldruin, line 44: Was this already explained somewhere else?

src/06-serial-communication/nix-tooling.md
```diff
@@ -19,42 +19,9 @@ the board; check the next section, which contains instructions for older
...
-> find it using `dmesg`, instead use `ls -l /dev | grep cu.usb` and adjust the following 
-> commands accordingly!
-
-``` console
-$ dmesg | grep -i tty
-(..)
-[  +0.000155] usb 3-2: FTDI USB Serial Device converter now attached to ttyUSB0
-```
-
-But what's this `ttyUSB0` thing? It's a file of course! Everything is a file in \*nix:
```
eldruin, line 36: I think this explanation is nice

src/06-serial-communication/nix-tooling.md
```diff
@@ -19,42 +19,9 @@ the board; check the next section, which contains instructions for older
...
-> **NOTE** if the permissions above is `crw-rw----`, the udev rules have not been set correctly
-> see [udev rules](../03-setup/linux.html#udev-rules)
-
-You can send out data by simply writing to this file:
-
-``` console
-$ echo 'Hello, world!' > /dev/ttyUSB0
-```
-
-You should see the TX (red) LED on the serial module blink, just once and very fast!
```
eldruin, line 52: Except for the different device name, this can still be done, right?

src/06-serial-communication/nix-tooling.md
```diff
@@ -97,8 +64,8 @@ baud rate to 115200. A text-based user interface (TUI) will pop out.
 </p>
 
 You can now send data using the keyboard! Go ahead and type something. Note that
-the TUI will *not* echo back what you type but, if you are using an external
-module, you *may* see some LED on the module blink with each keystroke.
+the TUI will *not* echo back what you type. If you pay attention to the yellow LED
```
eldruin, line 67:
```suggestion
the text UI will *not* echo back what you type. If you pay attention to the yellow LED
```

src/07-uart/README.md
```diff
@@ -0,0 +1,8 @@
+# UART
+
+The microcontroller has a peripheral called USART, which stands for Universal
+Asynchronous Receiver/Transmitter. This peripheral can be configured to work with
```
eldruin, line 4: UART/USART

eldruin (commented):
Looks good to me so far, thanks!
What I think it is a bit sad is getting rid of the registers chapter. I understand it is too much for this level, though. If we do get rid of the F3 book version I would prefer seeing that chapter as an appendix here or somehow make its way into the embedonomicon. I think it is well written and explains easily what goes on behind the scenes, which is so often left out or is explained in a more complicated way.

src/06-serial-communication/nix-tooling.md
hargoniX: In theory it's possible that this info might be wrong with the micro:bit so I removed it (I cannot check whether it's true since I dont own a Mac) but let's put it back in until proven wrong ^^

hargoniX: Actually reading through the old version again, this part of the document is named " ## Older revisions of the discovery board / external serial module" so it's probably irrelevant if the stuff before this chapter already works? Would be cool if someone with a Mac could verify this.

src/06-serial-communication/nix-tooling.md
hargoniX: This is again only for outdated revision of the board/the serial adapter, on my linux I can just call minicom onto the device with my permissions looking like what is described as "wrong" here:
```
λ ls -l /dev/ttyACM0
crw-rw----. 1 root plugdev 166, 0 Jan 21 11:56 /dev/ttyACM0
```
So I guess this is fine? At least as long as you are part of the plugdev group of course but that might also be a Fedora specific thing, not quite sure :/

src/06-serial-communication/nix-tooling.md
hargoniX: True, I'll add it back in!

src/06-serial-communication/nix-tooling.md
eldruin: I will check on it

src/06-serial-communication/nix-tooling.md
eldruin: I did not have to set up permissions either in ubuntu so I would agree to the removal.

src/07-uart/send-a-single-byte.md
```diff
@@ -0,0 +1,48 @@
...
+There is some parts we have already seen before but also lots of new stuff so lets focus on those.
+First things first, how do we know that we have to interact with P0_24 and P0_25 here? If you take a look
+at the micro:bit schematics page 5 again, you will notice that those two pins are also referred to as TGT_RX
+and TGT_TX. Not only that, they are also connected to our debug probe on page 2 so those are most likely the two
+pins we are looking for.
+
+Next up we can observe a new pin configuration we haven't seen before `p0.p0_25.into_floating_input()`, this just
+means that P0_25 is now an input, what exactly floating means is none of our concern right now. But how do we know
+we have to put our pins in this exact mode? If you look at the line below you can see that we construct an instance
+of `Pins` which is later passed on to our UART peripheral constructor (cts and rts are for more advanced UART features
```
eldruin, line 22:
```suggestion
of `Pins` which is later passed on to our UART peripheral constructor (`cts` and `rts` are for more advanced UART features
```
I'm guessing. Otherwise "cts" and "rts" in quotes I would say.

src/07-uart/send-a-single-byte.md
```diff
@@ -0,0 +1,48 @@
...
+we do not use here, hence they are set to None) and the [type signature](https://docs.rs/nrf51-hal/0.11.0/nrf51_hal/uart/struct.Pins.html)
```
eldruin, line 23:
```suggestion
we do not use here, hence they are set to `None`) and the [type signature](https://docs.rs/nrf51-hal/0.11.0/nrf51_hal/uart/struct.Pins.html)
```

src/07-uart/send-a-single-byte.md
```diff
@@ -0,0 +1,48 @@
...
+of `Pins` already tells us what to do.
+
+Next we construct our UART peripheral with `hal::Uart::new(p.UART0, pins, hal::uart::Parity::EXCLUDED, hal::uart::Baudrate::BAUD115200);`.
```
eldruin, line 26:
```suggestion
Next, we construct our UART peripheral with `hal::Uart::new(p.UART0, pins, hal::uart::Parity::EXCLUDED, hal::uart::Baudrate::BAUD115200);`.
```

src/07-uart/send-a-single-byte.md
```diff
@@ -0,0 +1,48 @@
...
+This function takes ownership of `p.UART0` and our `pins` so nobody else can mess with either the UART peripheral or our pins while
+we are using them. After that we pass two configuration options two the constructor, the baudrate (that one should be
```
eldruin, line 28:
```suggestion
we are using them. After that we pass two configuration options two the constructor: the baudrate (that one should be
```

src/07-uart/send-a-single-byte.md
```diff
@@ -0,0 +1,48 @@
...
+familiar) as well as an option called parity. Parity is a way that allows serial communication lines to check whether
```
eldruin, line 29:
```suggestion
familiar) as well as an option called "parity". Parity is a way that allows serial communication lines to check whether
```
Emphasis

src/07-uart/send-a-single-byte.md
```diff
@@ -0,0 +1,48 @@
...
+the data they received was corrupted on the line or not but we don't want to use that here so we simply exclude it.
```
eldruin, line 30:
```suggestion
the data they received was corrupted during transmission or not but we don't want to use that here so we simply exclude it.
```

src/07-uart/send-a-single-byte.md
```diff
@@ -0,0 +1,48 @@
...
+
+Last but not least we send our `X` via the newly created uart instance. The `block!` macro here is the `nb::block!`
```
eldruin, line 32:
```suggestion
Last but not least, we send our `X` via the newly created uart instance. The `block!` macro here is the `nb::block!`
```

src/07-uart/send-a-single-byte.md
```diff
@@ -0,0 +1,48 @@
...
+macro. `nb` is a (quoting from its description) "Minimal and reusable non-blocking I/O layer". It allows us to write
+code that can conduct hardware operations in the background while we go and do other work (non-blocking). However
```
eldruin, line 34:
```suggestion
code that can conduct hardware operations in the background while we go and do other work (non-blocking). However,
```

src/07-uart/send-a-single-byte.md
```diff
@@ -0,0 +1,48 @@
...
+
+Before flashing this you should make sure to start your minicom/PuTTY as the data we receive via our serial
+communication is not backed up or anything, we have to view it live. Once your serial monitor is up you can
+do a simple
+
+```
+$ cargo embed --release
+```
+
+And after the flashing you should see the character `X` show up on your minicom/PuTTY terminal, congrats!
```
eldruin, line 48:
```suggestion
And after the flashing is finished, you should see the character `X` show up on your minicom/PuTTY terminal, congrats!
```

eldruin (commented):
Great stuff! It reads quite nicely.
I just added some minor suggestions.

src/06-serial-communication/nix-tooling.md
```diff
@@ -1,60 +1,37 @@
 # \*nix tooling
 
-## Newer revisions of the discovery board
+## Newer revisions of the micro:bit board
 
-With newer revisions, if you connect the discovery board to your computer you
+With newer revisions, if you connect the micro:bit board to your computer you
```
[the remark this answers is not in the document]
hargoniX: Thinking about it, there arent actually new versions....I just replaced micro:bit with discovery here and the one sentence section was small enough for me to overlook :/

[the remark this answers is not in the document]
eldruin: Hmm, still there are 3 versions of the micro:bit: 1.3, 1.5 and 2.0. The question would be if that makes any difference in this regard.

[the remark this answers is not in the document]
hargoniX: iirc the new micro:bit 2.0 has an NRF52 CPU even? So it wont be compatible with this book at the moment :/ We're gonna have to look into that at some point. Regarding the difference between 1.3 and 1.5 I'm not quite sure, I know that i got a version 1.5 that behaves like described here.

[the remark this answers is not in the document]
eldruin: Yes, version 2.0 uses an nRF52833 but keeps the same motion sensor as version 1.5 (LSM303AGR). Version 1.3 has a different accelerometer and magnetometer.
Still, with regard to serial communication everything should be the same, I think. The example programs will have to differ a bit, though.
See [micro:bit hardware](https://tech.microbit.org/hardware/)

src/06-serial-communication/nix-tooling.md
hargoniX: THis mac behaviour isnt relevant anymore according to therealprof.

eldruin (approved):
Alright let's get this merged. Thank you for your work @hargoniX!

bors r+

src/06-serial-communication/nix-tooling.md
therealprof: Well, I said the `/dev/cu.*` isn't relevant. It might still be good to know how to find the device node... Not sure whether this is mentioned somewhere else. ;)

hargoniX:
That CI error is coming from a chapter in the later pats of the book referencing back to a chapter I now moved, any ideas what I should do regarding that? I guess I could just comment out the link or something?

hargoniX:
Since this is just a WIP branch anyways I'll go ahead and comment it out and remove it during a later chapter rewrite.

PR #285, opened 2021-01-16
Adds details of how to assess and fix an "Address already in use" error

liamhession:
On my machine (a Mac) i encountered this error every time i went to start up openocd according to the Verify the Installation instructions in chapter 3. For me, it was because some java process was running on port 4444. I went the Solution A route and killed that process.

For the typical rust-embedded user this troubleshooting step may be quite obvious. But when i did not see my particular error message in the troubleshooting section, i got a little lost wondering if any number of way less likely scenarios were affecting my device.

Because i don't have access to a Linux or Windows machine, i have not included instructions on how to find and kill processes using particular ports on those platforms. I thought of copy-pasting instructions from the internet, but i would not be able to verify those instructions work as described.

### DIFF
--- a/src/appendix/1-general-troubleshooting/README.md
+++ b/src/appendix/1-general-troubleshooting/README.md
@@ -137,6 +137,42 @@ Or, `itmdump` was called **after** the `monitor tpiu` was issued thus making
 - Then, launch `itmdump`
 - Then, launch the GDB session that executes the `monitor tpiu` command.
 
+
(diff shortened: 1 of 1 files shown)


eldruin (approved):
Sounds good to me. Although the solution is only provided for macOS, I think it can provide at least some guidance for other users. Maybe somebody else can enhance this in the future.
Thank you!

bors r+

Issue #284, opened 2021-01-10
7.2. (mis)Optimization : `cargo objdump` syntax broken


jerome-noel:
First instance appears like this :
```
$ cargo objdump --bin registers -- -d -no-show-raw-insn -print-imm-hex -source
```

when it should be 
```
$ cargo objdump --bin registers -- -d --no-show-raw-insn --print-imm-hex --source
```

e.g a second dashes are missing in the command argument after the middle `-- -d` .
The other few instances are also broken in the same way.

PR #281, opened 2020-12-30
07-registers: Fix cargo objdump example

gondolyr:
The last code block of the Type Safe Manipulation section contains
invalid flags because they are missing a dash. Thankfully cargo is
intelligent enough to provide correct suggestions.

Suggestion from cargo:

```console
llvm-objdump: Unknown command line argument '-no-show-raw-insn'.  Try: '/home/itme/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin/llvm-objdump --help'
llvm-objdump: Did you mean '--no-show-raw-insn'?
llvm-objdump: Unknown command line argument '-print-imm-hex'.  Try: '/home/itme/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin/llvm-objdump --help'
llvm-objdump: Did you mean '--print-imm-hex'?
```

PR #291, opened 2021-02-01
Fix specifying options in (mis)Optimization

bluk:
- The long argument name options specified to `objdump` require
  two dashes '--'.

Fixes https://github.com/rust-embedded/discovery/issues/284 .

### DIFF
--- a/src/07-registers/optimization.md
+++ b/src/07-registers/optimization.md
@@ -61,8 +61,8 @@ register, but the *release* (optimized) program only has one.
 We can check that using `objdump`:
 
 ``` console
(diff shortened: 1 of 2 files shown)


bluk:
Did a `git grep objdump` and seemed that Chapter 9.2 NOP also required a `--` in the options.

This is following up fixes from https://github.com/rust-embedded/discovery/pull/281 which fixed Chapter 7.5 Type safe manipulation.

eldruin (approved):
Sounds good to me, thanks!

bors r+

PR #297, opened 2021-02-08
Tweak testing cargo-binutils

winksaville:
While following the instructions for cargo-binutils in 03-setup
the `cargo size -- -version` caused an error when following the
instructions:

``` console
$ cargo size -- -version
error: Error during execution of `cargo metadata`: error: could not find `Cargo.toml` in `C:\Users\wink` or any parent directory
```

The cause of the error is that you're not told to  be in a rust project
directory.

The solution I've chosen is to create a rust bin project, cd into the
project, run it and then do the `cargo size -- -version`.

### DIFF
--- a/src/03-setup/README.md
+++ b/src/03-setup/README.md
@@ -90,13 +90,24 @@ itmdump 0.3.1
 ``` console
 $ rustup component add llvm-tools-preview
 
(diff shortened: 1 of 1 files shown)


eldruin (approved):
Nice catch. Thank you!

bors r+

Issue #289, opened 2021-02-01
Switch to GitHub Actions for CI


eldruin:
TODO:
- [ ] Switch
- [x] Fix publication issue about missing dateutil python package. See [https://travis-ci.org/github/rust-embedded/discovery/builds/756906140#L577]
```
+curl -Ls https://github.com/davisp/ghp-import/archive/master.tar.gz
+./ghp-import/ghp_import.py book
Traceback (most recent call last):
  File "./ghp-import/ghp_import.py", line 10, in <module>
    from dateutil import tz
ImportError: No module named dateutil
```
This problem is also present on https://github.com/rust-embedded/book/issues/280

PR #298, opened 2021-02-08
Fix installing dateutil since it is now a dependency of the GHP import script

eldruin:
This would at least make our publication step work again even though still on Travis CI.
See #289


adamgreig (approved):
Thanks!

bors merge

Issue #274, opened 2020-10-31
Newer revisions use LSM303AGR instead of LSM303DLHC


aditsachde:
Newer revisions of the discovery board use the `LSM303AGR` instead of the `LSM303DLHC`. The I2C address of the `LSM303AGR` chip is still the same but the registers to read are different. This makes it very confusing, as the code doesn't end up panicking, only returning zeros or garbage data.

eldruin:
Thanks for noticing. I have a WIP driver for the [LSM303AGR](https://github.com/eldruin/lsm303agr-rs) so you can at least play around with that.
It is actually the same accelerometer as the micro:bit board has, for which this book is being rewritten for (a deprecation/removal of the current book version is still uncertain since the F3 disco board seems available for purchase again)

aditsachde:
That's great news! Even with the discrepancies the book has, I found this book extremely helpful with dipping my toes into the embedded rust world, and updating the book will help countless more people.

Winnaries:
Additionally, to make the `LSM303AGR` works like in the section 14.5 of the discovery book, you have to change its system mode from default `idle` to `continuous` mode by modifying the first two bits of the register `CFG_REG_A_M` via the I2C as well. Otherwise, your `(x, y, z)` value would probably stuck at some values! Hope this help someone.

wbuck:
This is for anyone working through the book and you're stuck at the `I2C` section because the returned value from the `IRA_REG_M` register is `0`.

Here is the link for the [LSM303AGR](https://www.st.com/resource/en/datasheet/lsm303agr.pdf) data sheet. As noted above, newer board revisions use this `IC` instead of the `LSM303DLHC`.

For a quick test the `WHO_AM_I_M` register (address `0x4F`) of the magnetometer can be used to quickly query the device via I2C.

If successfully you should see `64` or `0b01000000`

So the code from the book becomes:

    const MAGNETOMETER: u8 = 0b001_1110;
    // Addresses of the magnetometer's registers.
    const WHO_AM_I_M: u8 = 0x4F;

    #[entry]
    fn main() -> ! {
        let (i2c1, _delay, mut itm) = aux14::init();

        // Stage 1: Send the address of the register we want to read to the
        // magnetometer
        {
            // Broadcast START
            // Broadcast the MAGNETOMETER address with the R/W bit set to Write
            i2c1.cr2.write(|w| {
                w.start().set_bit();
                w.sadd1().bits(MAGNETOMETER);
[...]

PR #300, opened 2021-02-11
add two notes flagging lsm303agr

waalge:
In response to  https://github.com/rust-embedded/discovery/issues/274#issuecomment-777245460

Please edit at leisure / suggest anything


eldruin (approved):
This can always be further improved but it should help users already.
Thank you!

bors r+

PR #304, opened 2021-02-28
Update-03-setting-up-a-dev-env

winksaville:
Update `*.md` files so that copy-to-clipboard button in code blocks
works so that commands can be copied and pasted into terminal windows.

Some reorganization of the text for clarity.


adamgreig (approved):
Thank you for this PR and all your recent work!

bors merge

PR #307, opened 2021-03-03
DO NOT Merge - Test shard cargo config and openocd.gdb

winksaville:
Created src/.cargo/config.toml which references `../openocd.gdb` which
is the src/openocd.gdb.

Create src/openocd.gdb which is the most general openocd.gdb defined by
any of the chapters

Tested 05, 06 and 07 locally and this seems to work and also ran
TRAVIS_BRANCH=hello bash ci/script.sh which seemed to work.

Changed src/05-led-roulette/.cargo/config to config.tomol and have it
be an "empty" cargo config by commenting out the assignment statements.
This represents one possibility. Also, we need to decide what to do
about local openocd.gdb files.

Renamed all .cargo/config to ./cargo/config.original so as to "prove"
they are not needed locally.


winksaville:
I'd like to propose we add `embedded-discover/src/.cargo/config.toml` and `embedded-discovery/src/openocd.gdb` this will make it easier for the user. As it stands now if the users system doesn't have `arm-none-eabi-gdb` installed they have to change every  `.cargo/config` file in each of the chapters. By moving `.carog/config.toml` and `openocd.gdb` to `src/` it only has to be changed in one place.

If this is acceptable I'll chnge 05-roulette/flash-it to direct the user to update `embedded-discover/src/.cargo/config.toml` to the appropriate `runner =` line in the "Update .cargo/config" section.

The other thing to do is decide is what to do with the `src/*/.cargo/config` and `src/*/openocd.gdb`. AFAICT we can just remove them. We could also just "comment out" the contents or ...

Please give me feed back!

eldruin (approved):
Thank you!
In my opinion, one central file seems simpler for the reasons you listed. I would also just remove all other files since we have better ways for people to start a new project than to copy a small subproject from this book.

PS: In GitHub you can mark a PR which is not ready to be merged as a "draft PR" and later as ready when it is.

eldruin:
This was merged in #308

PR #308, opened 2021-03-05
Add shareable instances of config.toml and openocd.gdb

winksaville:
Instead of every chapter having its own copy of the cargo configuration
and openocd.gdb files this change creates a shared set. This will make it
much easier for the user of the discovery book to handle the situation
where their Arm gdb is NOT arm-none-eabi-gdb. As only the shared copy
of .cargo/config.toml will have to be modified.

Also src/05-led-roulette is updated to take advantage of this and I
will go through each of the other chapters changing them to use
the shared set.

I also needed to comment out the rustflags variable in all of the config
files as I'd get an error executing ci/script.sh:

  ry.x:5: region 'FLASH' already defined
          >>>   FLASH : ORIGIN = 0x08000000, LENGTH = 256K
          >>>

PR #309, opened 2021-03-05
Update 06 hello world to use shared cargo config

winksaville:
This is dependent on #308 but there is a reduction in number of lines for 06-hello-world, usually a positive :)


src/.cargo/config.toml
```diff
@@ -1,8 +1,11 @@
 [target.thumbv7em-none-eabihf]
```
eldruin, line 1: This file should be removed, right? It will clash with #308

eldruin (commented):
Looks good to me, thanks!

src/.cargo/config.toml
winksaville: If I'm not mistaken this git showing the deletion of `.cargo/config` as a "rename", it's confused :) I think when #308 is merged this may change. In anycase there is no `src/06-hello-world/.cargo/config` in my version of `src/06-hello-world`.

eldruin (approved):
bors r+

src/.cargo/config.toml
eldruin: Hmm, GitHub is showing a merge conflict with the `src/06-hello-world/.cargo/config` file. Could you rebase this to master?

winksaville:
fingers crossed :)

eldruin (approved):
Let's try.
bors r+

winksaville:
Weird, it looks like the  "staging" step took 27+ minutes with the last two PRs. Where as before it was taking half that time and even earlier staging runs were in the 6 minute range. Is there something about the common ../.cargo/config.toml that is slowing things down?

eldruin:
I would tip this is just the throttling on Travis. Sometimes it hangs forever in the schedule queue.

PR #311, opened 2021-03-10
Swap to GHA

adamgreig:
:crossed_fingers:


adamgreig:
Ok, that wasn't _too_ bad. Ready for review.

PR #328, opened 2021-04-23
Fix breaking change in  cortex-m 0.5.11.

wpd:
Fix it by sticking with cortex-m 0.5.6.

See discussion at #327.  This is a much simpler (and certainly more hackish) solution than is proposed in #267.  But it doesn't require `unsafe` code nor a custom `memory.x` file, and might fill the gap until the book is rewritten.

### DIFF
--- a/src/11-usart/auxiliary/Cargo.toml
+++ b/src/11-usart/auxiliary/Cargo.toml
@@ -5,7 +5,7 @@ name = "aux11"
 version = "0.1.0"
 
 [dependencies]
-cortex-m = "0.5.6"
+cortex-m = "=0.5.6" # 0.5.11 introduces a breaking change.  Use 0.5.6, since we know it works for this example
 cortex-m-rt = "0.6.3"
 panic-itm = "0.4.0"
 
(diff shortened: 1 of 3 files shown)


adamgreig (dismissed):
Thanks, this is a really simple fix that will tide us over nicely, good idea.

bors r+

adamgreig:
Ugh, because all the chapters are in a workspace, Cargo sees the new `=0.5.6` from chapter 11 when trying to build chapter 15, which then tries to use both 0.5.6 and 0.7 and fails out. Could you please try also setting `=0.5.6` in chapters 15 and 16?

wpd:
I tried changing `cortex-m` to `=0.5.6` throughout the project, and that resulted in thing like `05-led-roulette` failing to build.  So then I tried changing it in just chapter 15; running `cargo build --target=thumbv7em-none-eabihf` in the top level directory; seeing that fail for chapter 14; fixing the chapter 14 Cargo.toml file; and seeing the top level build (all?) work.  Hopefully this will pass the CI tests.

--wpd

adamgreig (approved):
Thanks, let's try it!

bors merge

adamgreig:
bors retry

PR #337, opened 2021-05-11
Fixed typos regarding `.cargo/config.toml`

rdelfin:
The file keeps on being referred to as either `cargo/config.toml` or `.cargo/config.toml`. This PR corrects all of them to `.cargo/config.toml`

### DIFF
--- a/src/05-led-roulette/flash-it.md
+++ b/src/05-led-roulette/flash-it.md
@@ -143,11 +143,11 @@ that port.
 ## Update ../.cargo/config.toml
 
 Now that you've successfully determined which debugger you need to use
-we need to change `../cargo/config.toml` so that `cargo run` command will succeed.
+we need to change `../.cargo/config.toml` so that `cargo run` command will succeed.
 Note: `cargo` is the rust package manager and you can read about it
 [here](https://doc.rust-lang.org/cargo/).
 
-Get back to the terminal prompt and look at `../cargo/config.toml`:
+Get back to the terminal prompt and look at `../.cargo/config.toml`:
 ``` console
 ~/embedded-discovery/src/05-led-roulette
 $ cat ../.cargo/config.toml


adamgreig (approved):
Thanks!

bors merge

adamgreig:
bors retry

rdelfin:
The wonders of CI

sirhcel:
> Huh, that's more worrying. Not your fault though, will investigate.

It looks like depending on two panic handlers (`panic-halt` and `panic-itm`) works fine when compiling the example code (as it actually only uses `panic-itm`). But it looks like `cargo doc` does not take the actual usage into account and fails due to the duplicated `panic_impl`.

Removing the dependencies on `panic-halt` fixes building the documentation for me (https://github.com/sirhcel/discovery/commit/3e2a1807d294ac5dced878c53b1670b162750fa3).

sirhcel:
@adamgreig and @rdelfin is there anything speaking against removing `panic-halt` to fix building the book? I don't see a reason why the dependency to `panic-halt` was there in the first place.

In case you have no objections how to proceed? Should I create a PR against your branch @rdelfin? Our would you simply pick the commit?

adamgreig:
If that's enough to get it working please do go ahead with a PR, probably against master is easiest since we can merge that first and then when we retry these other PRs, they will be tested against the merged version.

rdelfin:
I'd be glad to rebase on top of that change after it merges, given that's probably the easiest option!

sirhcel:
I'm sorry, but it looks like I had a mental blackout late at night when writing https://github.com/rust-embedded/discovery/pull/337#issuecomment-848918792. My proposal fixes `cargo doc` but none of the actual issues discussed here. I'm really sorry for bugging you! :(

#267 and #341 already propose actually suitable changes and you already pointed out in the latter that there is some real work ahead @adamgreig. I will have a look if I could contribute something to these PRs.

eldruin (approved):
This should now work after the latest fixes.

bors r+

PR #343, opened 2021-05-31
Clean up warning on unused result

sirhcel:
I would prefer the example code not to contain unused results. This cleans up the actually one and only warning on an unused result.
```bash
$ cargo build --target thumbv7em-none-eabihf
[...]
warning: unused `Result` that must be used
  --> src/16-punch-o-meter/auxiliary/src/lib.rs:38:5
   |
38 |     nss.set_high();
   |     ^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_must_use)]` on by default
   = note: this `Result` may be an `Err` variant, which should be handled

warning: 1 warning emitted
[...]
```


adamgreig (approved):
Thanks!

bors merge

Issue #314, opened 2021-03-20
cargo doc => error: duplicate lang item in crate `panic_halt`: `panic_impl`.


ereichert:
I cloned the repo today and ran cargo doc on master in the /src/05-led-roulette directory.

`==> cargo doc
 Documenting aux5 v0.2.0 (/Users/ereichert/dev/tmp/discovery/src/05-led-roulette/auxiliary)
error: duplicate lang item in crate `panic_halt`: `panic_impl`.
  |
  = note: the lang item is first defined in crate `panic_itm` (which `aux5` depends on)
  = note: first definition in `panic_itm` loaded from /Users/ereichert/dev/tmp/discovery/target/thumbv7em-none-eabihf/debug/deps/libpanic_itm-eaa2b2ea5c1bfc9c.rmeta
  = note: second definition in `panic_halt` loaded from /Users/ereichert/dev/tmp/discovery/target/thumbv7em-none-eabihf/debug/deps/libpanic_halt-995f20f96388dea2.rmeta

error: aborting due to previous error

error: could not document `aux5`

Caused by:
[...]

PR #344, opened 2021-05-31
Fix building docs and add CI check

sirhcel:
Declaring `panic-halt` as a depencency works fine for building the example code but it makes `cargo doc` fail due to the duplicate lang item `panic_impl`.

This issue got already reported with https://github.com/rust-embedded/discovery/issues/314 and gets fixed by this PR by simply removing the unused `panic-halt` handler.


sirhcel:
Going through the actual issues, this PR will also fix #322.

sirhcel:
@eldruin suggested to check docs with CI as well. So I added CI checks as an afterthought.

eldruin (approved):
Nice! looks good to me. Thank you!

bors r+

PR #345, opened 2021-05-31
LED Challenge section instructs edit of wrong file.

ereichert:
Corrected the file to be edited to add the GDB configuration in the LED challenge section.


ereichert:
Someone else already got to this.

PR #349, opened 2021-06-13
fix typos in openocd.gdb comments

Jonahss:


Jonahss:
Loving the Discovery book so far! Setup was painless.

src/openocd.gdb
```diff
@@ -29,7 +29,7 @@ break DefaultHandler
 # Set a breakpiont at HardFault
 break HardFault
 
-# Continue running and unill we hit the main breakpoint
+# Continue running and until we hit the main breakpoint
```
adamgreig, line 32:
```suggestion
# Continue running until we hit the main breakpoint
```

adamgreig (commented):
Thanks! Looks good but I think we can get rid of "and" here too?

Jonahss:
@adamgreig included your suggestion. Squashed into one commit.

adamgreig (approved):
thanks!

bors merge

PR #352, opened 2021-06-21
DOC Minor fixes

NitinSaxenait:
This is a minor change in the RTRM Section 7.1 which makes it easy to find the BSRR register.
The way we can reach to BSRR register and it's offset value 0x18.

### DIFF
--- a/src/07-registers/rtrm.md
+++ b/src/07-registers/rtrm.md
@@ -37,7 +37,7 @@ peripheral, that table is in:
 
 > Section 11.4.12 GPIO register map - Page 243
 
-'BSRR' is the register which we will be using to set/reset. It's offset value is '0x18' from the base address 
+'BSRR' is the register which we will be using to set/reset. Its offset value is '0x18' from the base address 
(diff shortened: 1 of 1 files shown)


src/07-registers/rtrm.md
```diff
@@ -37,8 +37,9 @@ peripheral, that table is in:
 
 > Section 11.4.12 GPIO register map - Page 243
 
-We are interested in the register that's at an offset of `0x18` from the base address of the `GPIOE`
-peripheral. According to the table, that would be the register `BSRR`.
+'BSRR' is the register which we will be using to set/reset. It's offset value is '0x18' from the base address
```
eldruin, line 40:
```suggestion
'BSRR' is the register which we will be using to set/reset. Its offset value is '0x18' from the base address 
```

eldruin (commented):
Other than the typo looks good to me. thanks!

eldruin (approved):
Looks good to me. Thanks!

bors r+

PR #356, opened 2021-06-22
Doc minor fixes for command cargo run --release

NitinSaxenait:
### This suggests us to use another debugger in register/.cargo/config.


eldruin:
Thanks! could you solve the conflicts here?
(In the future, it would be better to make the PRs independent, as this includes changes from #417)

PR #357, opened 2021-06-22
Switch back to MonoTimer from HAL

sirhcel:
According to https://github.com/stm32-rs/stm32f3xx-hal/pull/247 it got removed accidentially and has been resurrected. There is no need for the local copy from here.

- [ ] Bump BSP and HAL once the latter gets a new release including https://github.com/stm32-rs/stm32f3xx-hal/pull/247
- [ ] Rebase to latest master at this point in time


sirhcel:
A fixed HAL got released with [0.8.0](https://github.com/stm32-rs/stm32f3xx-hal/releases/tag/v0.8.0) meanwhile. So https://github.com/rubberduck203/stm32f3-discovery/pull/44 is on the way for updating the BSP.

BartMassey:
Not sure after all this time of the relevance? Would probably need to be updated for HAL 1.x

sirhcel:
You are absolutely right. This draft PR has been dry-aging and with https://github.com/rust-embedded/wg/pull/759 on the horizon this looks like too much effort for restoring nostalgia.

PR #362, opened 2021-07-11
Streamline LED roulette example

sirhcel:
This PR brings the LED roulette to up-to-date dependencies and factors out the examples, so that they could be tested in CI.

I wanted to update all examples at once in the first place. But this did not get to the pace I wanted, so I will attempt this one after another.


adamgreig (approved):
Great, thanks!

bors merge

Issue #364, opened 2021-07-11
Sec 3.1 - Fedora package does not exist


anthonyjmartinez:
`arm-none-eabi-gdb` appears to have ceased existing in default (or RPMFusion) repositories somewhere around F30, recent discussion around the topic seems to indicate it's not even necessary: https://bugzilla.redhat.com/show_bug.cgi?id=1859627#c23

sirhcel:
Thank you for pointing that out @anthonyjmartinez! I created PR #367 for guiding Fedora users to use GDB from a pre-built toolchain from ARM instead.

anthonyjmartinez:
@sirhcel thanks for the speedy update.

I think it's worth noting that this JustWorks™ with the normal `gdb` from Fedora 33:

OpenOCD gave a warning that `stlink.cfg` should be used (it works either way though):
```
[user@projects ~]$ sudo openocd -f /usr/share/openocd/scripts/interface/stlink.cfg -f /usr/share/openocd/scripts/target/stm32f3x.cfg 
Open On-Chip Debugger 0.11.0
Licensed under GNU GPL v2
For bug reports, read
        http://openocd.org/doc/doxygen/bugs.html
Info : auto-selecting first available session transport "hla_swd". To override use 'transport select <transport>'.
Info : The selected transport took over low-level target control. The results might differ compared to plain JTAG/SWD
Info : Listening on port 6666 for tcl connections
Info : Listening on port 4444 for telnet connections
Info : clock speed 1000 kHz
Info : STLINK V2J37M26 (API v2) VID:PID 0483:374B
Info : Target voltage: 2.905338
Info : stm32f3x.cpu: hardware has 6 breakpoints, 4 watchpoints
Info : starting gdb server for stm32f3x.cpu on 3333
Info : Listening on port 3333 for gdb connections
Info : accepting 'gdb' connection on tcp/3333
Info : device id = 0x10036422
Info : flash size = 256kbytes
```

GDB says to use `extend-remote` instead of `remote`
```
[user@projects 05-led-roulette]$ gdb -q -ex "target extended-remote :3333" target/thumbv7em-none-eabihf/debug/led-roulette
Reading symbols from target/thumbv7em-none-eabihf/debug/led-roulette...
Remote debugging using :3333
warning: Remote gdbserver does not support determining executable automatically.
RHEL <=6.8 and <=7.2 versions of gdbserver do not support such automatic executable detection.
The following versions of gdbserver support it:
[...]

PR #367, opened 2021-07-11
Fix instructions for installting GDB on Fedora

sirhcel:
The package `arm-none-eabi-gdb` is no longer provided by Fedora as stated in #364. So let's guide Fedora uses to _Other distros_ then and use a pre-built toolchain provided by ARM. This is also the recommendation from the [issue at Fedora](https://bugzilla.redhat.com/show_bug.cgi?id=1859627#c26).

### DIFF
--- a/src/03-setup/linux.md
+++ b/src/03-setup/linux.md
@@ -42,12 +42,10 @@ sudo apt-get install \
 
 ### Fedora 23 or newer
 
-> **NOTE** `arm-none-eabi-gdb` is the GDB command you'll use to debug your ARM
-> Cortex-M programs
+Install GDB as described in [Other distros](#other-distros) and the following packages:
 
 ``` console
 sudo dnf install \
-  arm-none-eabi-gdb \
   minicom \
   openocd
 ```


adamgreig (approved):
Thanks!

I suspect Fedora users can probably use gdb-multiarch as well, but the new instructions here should also work fine. Perhaps we should link to a newer version though...

bors merge

sirhcel:
> I suspect Fedora users can probably use gdb-multiarch as well, but the new instructions here should also work fine. 

I had a quick look at https://src.fedoraproject.org/projects/rpms/* but it seems that this package is not available.

> Perhaps we should link to a newer version though...

Do you mean the version of the toolchain? _10-2020-q4-major_ is still the latest release according to their website.

adamgreig:
Huh, I guess no further work required then! I assumed there would have been a 2021 release of the GCC toolchain by now but you're right, there's not, so nothing to update. Thanks for checking!

hargoniX:
I have been working with Fedora's gdb for the longest time and just using the default package just works :tm: for me:
```
λ gdb target/thumbv7em-none-eabihf/debug/examples/adc
GNU gdb (GDB) Fedora 10.2-3.fc34
<snip>
Reading symbols from target/thumbv7em-none-eabihf/debug/examples/adc...
(gdb) l main
13      use stm32h7xx_hal::{adc, delay::Delay, pac, prelude::*, rcc::rec::AdcClkSel};
14
15      #[macro_use]
16      mod utilities;
17
18      #[entry]
19      fn main() -> ! {
20          utilities::logger::init();
21          let cp = cortex_m::Peripherals::take().unwrap();
22          let dp = pac::Peripherals::take().unwrap();
(gdb)
```
My only GDB related packages are:
```
λ dnf list --installed | grep gdb
gap-pkg-liealgdb.noarch                           2.2.1-4.fc34                           @fedora
gcc-gdb-plugin.x86_64                             11.1.1-3.fc34                          @updates
gdb.x86_64                                        10.2-3.fc34                            @updates
gdb-gdbserver.x86_64                              10.2-3.fc34                            @updates
gdb-headless.x86_64                               10.2-3.fc34                            @updates
gdbm-libs.x86_64                                  1:1.19-2.fc34                          @fedora
wireless-regdb.noarch                             2020.11.20-2.fc34                      @fedora
```
so it appears to me default fedora gdb just is multiarch already?

PR #371, opened 2021-07-14
Rewrite of chapter 2 for microbit v2

hargoniX:

PR #375, opened 2021-07-15
Chapter 4 rewrite micro:bit v2

hargoniX:
Note that this PR is based on #371 for the new micro:bit v2 image so that one should be merged before this.


eldruin (approved):
Looks good to me, thank you!

bors r+

PR #377, opened 2021-07-16
8.3: Use `write` instead of `modify`

WaffleLapkin:
Since the read value is ignored, using `modify` seems redundant. 
Also, `modify` isn't mentioned in the text prior this chapter, while `write` is explained.

### DIFF
--- a/src/08-leds-again/the-solution.md
+++ b/src/08-leds-again/the-solution.md
@@ -12,10 +12,10 @@ fn main() -> ! {
     let (gpioe, rcc) = aux8::init();
 
     // enable the GPIOE peripheral
-    rcc.ahbenr.modify(|_, w| w.iopeen().set_bit());
+    rcc.ahbenr.write(|w| w.iopeen().set_bit());
 
     // configure the pins as outputs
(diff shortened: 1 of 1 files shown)


jamesmunns (approved):
Looks good, thanks!

bors r+

adamgreig:
> Since the read value is ignored, using modify seems redundant.

Just to note, you use `modify` even if you're ignoring the read value to ensure it doesn't overwrite other fields in the process, since all fields you don't explicitly specify are set to their reset value when using `write` (but are kept at their current value when using `modify`).

In this case I think writing those registers to their reset values is OK because we haven't written anything else to them anyway, but it's worth being aware of, since it's a common mistake to use `write` and therefore blow away earlier changes to other fields which you wanted to keep.

For example:

```rust
rcc.ahb1enr.write(|w| w.gpioaen().enabled());
rcc.ahb1enr.write(|w| w.gpioben().enabled());
```

would leave _only_ GPIOB enabled, because the second write overrules the first. Using `modify` wouldn't have that issue.

PR #385, opened 2021-07-30
solving lsm303dlhc issue using lsm303agr

NitinSaxenait:
This basically use the lsm303agr package instead of lsm303dlhc which run the example of [LED compass](https://docs.rust-embedded.org/discovery/15-led-compass/index.html).


eldruin:
Thank you for this proposal. The main problem against it would be that this would break compatibility with older board revisions which do have an LSM303DLHC.

NitinSaxenait:
> Thank you for this proposal. The main problem against it would be that this would break compatibility with older board revisions which do have an LSM303DLHC.

NitinSaxenait:
> Thank you for this proposal. The main problem against it would be that this would break compatibility with older board revisions which do have an LSM303DLHC.

can we provide this particular solution for lsm303agr using a link in the doc? This can really help people to get the solution for compass.

eldruin:
That would definitely help people until we have a more comprehensive solution.

NitinSaxenait:
> That would definitely help people until we have a more comprehensive solution.

Thanks, i will provide my repo link for this particular solution in the doc so that we can help people carry on with lsm303agr

PR #387, opened 2021-08-05
use lsm303agr instead of lsm303dlhc in chapter compass

NitinSaxenait:
This basically provides the link to implement the compass code using the lsm303agr for the newer board. This removes the conflict until a better solution comes.


NitinSaxenait:
Created a new pull request after having a short discussion with @jamesmunns  . Can we help people using this until we get a better solution?

PR #390, opened 2021-08-14
Rewrite ch07 micro:bit v2

hargoniX:
Note that this again introduces a git dependency because we have to wait for some upstream releasese here.


hargoniX:
Rather weird it is trying to build the chapter 7 with build-chapter-old when the bors.toml i commited says to build it with build-chapter-microbit...

src/07-uart/README.md
```diff
@@ -4,5 +4,17 @@ The microcontroller has a peripheral called UART, which stands for Universal
 Asynchronous Receiver/Transmitter. This peripheral can be configured to work with
 several communication protocols like the serial communication protocol.
 
+
 Throughout this chapter, we'll use serial communication to exchange information between the
 microcontroller and your computer.
+
+> **NOTE** that on the micro:bit v2 we will use the so called UARTE peripheral which behaves
+> just like a regular UART, except that the HAL has to talk to it in a different way.
+> However this will of course not be our concern.
```
eldruin, line 13:
```suggestion
> However, this will of course not be our concern.
```

src/07-uart/README.md
```diff
@@ -4,5 +4,17 @@ The microcontroller has a peripheral called UART, which stands for Universal
...
+
+## Setup
+As always from now on you will have to modify the Embed.toml to match your micro:bit version:
```
eldruin, line 16:
```suggestion
As always from now on you will have to modify the `Embed.toml` to match your micro:bit version:
```

src/07-uart/naive-approch-write.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+While this is a perfectly valid implementation, at some point
+you might want to have all the nice perks of `print!` such
+as argument formatting and so on. If you are wondering how to do that, read on.
+
+## `write!` and `core::fmt::Write`
+The `core::fmt::Write` trait allows us to use any struct that implements
+it in basically the same way as we use `print!` in the `std` world.
+In this case the `Uart` struct from the `nrf` HAL does implement `core::fmt::Write`
```
eldruin, line 76:
```suggestion
In this case, the `Uart` struct from the `nrf` HAL does implement `core::fmt::Write`
```

src/07-uart/send-a-single-byte.md
```diff
@@ -9,40 +9,56 @@ In order to do that we will use the following snippet (this one is already in `0
...
-means that `P0_25` is now an input. What exactly floating means is none of our concern right now. But how do we know
-we have to put our pins in this exact mode? If you look at the line below you can see that we construct an instance
-of `Pins` which is later passed on to our UART peripheral constructor (`cts` and `rts` are for more advanced UART features
-we do not use here, hence they are set to `None`) and the [type signature](https://docs.rs/nrf51-hal/0.11.0/nrf51_hal/uart/struct.Pins.html)
-of `Pins` already tells us what to do.
-
-Next, we construct our UART peripheral with `hal::Uart::new(p.UART0, pins, hal::uart::Parity::EXCLUDED, hal::uart::Baudrate::BAUD115200);`.
-This function takes ownership of `p.UART0` and our `pins` so nobody else can mess with either the UART peripheral or our pins while
+You will also have noticed that this is the first time we are including some code that is not from a library,
+namely the `serial_setup` module. Its only purpose is to provide a nice and usable wrapper around the UARTE
```
eldruin, line 17:
```suggestion
namely the `serial_setup` module. Its only purpose is to provide a nice wrapper around the UARTE
```
The implication that without this wrapper the UARTE would be somewhat unusable seems harsh.

src/07-uart/send-a-single-byte.md
```diff
@@ -9,40 +9,56 @@ In order to do that we will use the following snippet (this one is already in `0
...
+so we can use it the exact same way as the UART via the [embedded_hal::serial] traits. If you want you can
```
eldruin, line 18:
```suggestion
so we can use it the exact same way as the UART via the [embedded_hal::serial] traits. If you want, you can
```

src/07-uart/send-a-single-byte.md
```diff
@@ -9,40 +9,56 @@ In order to do that we will use the following snippet (this one is already in `0
...
+check out what exactly the module does, but it is not required to understand this chapter in general.
+
+[embedded_hal::serial]: https://docs.rs/embedded-hal/0.2.6/embedded_hal/serial/index.html
+
+Apart from those differences the initilization procedures for the UART and the UARTE are quite similar so we'll
```
eldruin, line 23:
```suggestion
Apart from those differences, the initialization procedures for the UART and the UARTE are quite similar so we'll
```

src/07-uart/send-a-single-byte.md
```diff
@@ -9,40 +9,56 @@ In order to do that we will use the following snippet (this one is already in `0
...
+Then we wrap it up in the `UartePort` type so we can use it the same way as the micro:bit v1's `serial`.
 
-Last but not least, we send our `X` via the newly created uart instance. The `block!` macro here is the `nb::block!`
+After the initilization, we send our `X` via the newly created uart instance. The `block!` macro here is the `nb::block!`
 macro. `nb` is a (quoting from its description) "Minimal and reusable non-blocking I/O layer". It allows us to write
 code that can conduct hardware operations in the background while we go and do other work (non-blocking). However,
 in this and many other cases we have no interest in doing some other work so we just call `block!` which will wait until
 the I/O operation is done and has either succeeded or failed and then continue execution normally.
 
+Last but not least we `flush()` the serial port. This is because an implementor of the `embedded-hal::serial` traits may
```
eldruin, line 46:
```suggestion
Last but not least, we `flush()` the serial port. This is because an implementor of the `embedded-hal::serial` traits may
```

eldruin (commented):
Great work! Just a few nitpicks left.

eldruin (approved):
Thank you for your work!
bors r+

hargoniX:
Well sicne i changed the workflow for all i guess this was destined to blow up. We'll have to merge them one after the other.

PR #392, opened 2021-08-18
code to use lsm303agr in chapter LED compass

NitinSaxenait:
### What this implement/fix?

This add the code link to use lsm303agr package instead of lsm303dlhc for newer boards.

This will help people to get a temporary solution until some real fix happens.


eldruin:
It would be better to include the code directly here, rather than linking to an external repository.
Looking through the linked code, it also be better if the whole thing was just one file, rather than splitting initialization and so on up, since that would be friendlier to newcomers.

PR #394, opened 2021-08-21
Update the troubleshooting section for micro:bit

hargoniX:
This mostly just deletes stuff since its related to OpenOCD which we don't have to use anymore with this rewrite (\o/).
I haven't come across any issues with this (that I haven't fixed upstream already) apart from the two things mentioned in the
README. I would expect that this file might grow over time though.

### DIFF
--- a/src/appendix/1-general-troubleshooting/README.md
+++ b/src/appendix/1-general-troubleshooting/README.md
@@ -1,129 +1,17 @@
 # General troubleshooting
 
-## OpenOCD problems
+## `cargo-embed` problems
+Most `cargo-embed` problems are either related to not having installed the `udev`
+rules properly or having selected the wrong chip configuration in `Embed.toml` so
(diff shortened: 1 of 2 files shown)


src/appendix/1-general-troubleshooting/README.md
```diff
@@ -1,129 +1,17 @@
 # General troubleshooting
 
-## OpenOCD problems
+## `cargo-embed` problems
+Most `cargo-embed` problems are either related to not having installed the `udev`
+rules properly or having selected the wrong chip configuration in `Embed.toml` so
```
eldruin, line 5:
```suggestion
rules properly (on Linux) or having selected the wrong chip configuration in `Embed.toml` so
```

eldruin (commented):
Looks good to me!

eldruin (approved):
bors r+

PR #395, opened 2021-08-22
Rewrite ch09

hargoniX:
This PR is not quite done yet it's still missing magnitutde and calibration. However there is a thing that is confusing me. Both take 1 and 2 work perfectly on the micro:bit v2. On my micro:bit v1 however the X value of the magnetometer always stays negative...is this a mistake from my side or do I have to calibrate it?


src/09-led-compass/take-1.md
```diff
@@ -0,0 +1,88 @@
+# Take 1
+
+What's the simplest way in which we can implement the LED compass? Even if it's not perfect.
```
eldruin, line 3:
```suggestion
What's the simplest way in which we can implement the LED compass, even if it's not perfect?
```

src/09-led-compass/take-1.md
```diff
@@ -0,0 +1,88 @@
...
+
+For starters, we'd only care about the X and Y components of the magnetic field because when you
+look at a compass you always hold it in horizontal position thus the compass is in the XY plane.
```
eldruin, line 6:
```suggestion
look at a compass you always hold it in horizontal position and thus the compass is in the XY plane.
```

src/09-led-compass/take-1.md
```diff
@@ -0,0 +1,88 @@
...
+how the quadrant changes whenever we point in another direction.
+
+After experimenting a bit we can find out that if we point the micro:bit in e.g. south east direction,
+both the X and the Y component are always positive. Based on this information you should be able to
+figure out which direction the other quadrants represent.
+
+Once you figured out the relation between quadrant and direction you should be able to
+complete the template from below.
+
+``` rust
```
eldruin, line 24: Can we just include the code from the file here?

eldruin (commented):
Great stuff! just a few nitpicks.
Have you found out more about your accelerometer sign difference? it does not make much sense to me, since both board versions have the same chip and from visual comparison on my side, they seem to be soldered in the same orientation.

hargoniX:
I don't think its related to orientation or something...I turned 360 degrees around with the micro:bit v1 and the x values kept staying always negative which does not make sense at all...something weird is going on with the chip here.

caemor:
> I don't think its related to orientation or something...I turned 360 degrees around with the micro:bit v1 and the x values kept staying always negative which does not make sense at all...something weird is going on with the chip here.

Depending on the version of v1 you are using you might have a different magnetometer, although most likely your program shouldn't work at all in that case.

But since V1.3 is using a Freescale MAG3110 ( https://tech.microbit.org/hardware/1-3-revision/#magnetometer ), we should maybe make it clear that only v1.5 and v2 are supported here or that another feature needs to be enabled to use the mag3110 lib instead?

hargoniX:
The main issue here is with the calibration of the magnetometer on the micro:bit, depending on how its configured per default the values it outputs can wildly differ from other once which is why it should be properly calibrated in your current location before using it. I did that in MicroPython with the default micro:bit stuff and that got me quite accurate results so next step is to port that to Rust which is what I'm on at the moment.

hargoniX:
The code is now in a working state (along with the calibration)

There is one thing I am not quite sure yet, the magnitude of the magnetig field. The docs of the lsm303 crate as well as its datasheet agree that the data being outputted is in milli gauss so so the magnitude of the field vector should be in milli gauss as well. However the values are too large by that by the factor of 100 (which is exactly the conversion factor from nano tesla to milli gauss). The micro:bit original driver code notes that they calculate the magnitude in nano tesla and trying it out on my microbit they produce almost identical values so it would seem the values are being measured in nano tesla?

Since I more or less blindly took over the micro:bit calibration code maybe the conversion is coded somehwere in there as well, I am not 100% certain about that yet.

It would be very nice if someone could verify I translated the calibration code correctly and also clear up the unit question regarding nano tesla or milli gauss.

Apart from that (minus the merge conflicts) this should be ready to review and then merge.

src/09-led-compass/magnitude.md
```diff
@@ -1,67 +1,85 @@
...
-
-So, what we need to do is divide the X, Y and Z values that the sensor outputs by its corresponding
-*gain*. Then, we'll have the X, Y and Z components of the magnetic field in Gauss.
-
-With some extra math we can retrieve the magnitude of the magnetic field from its X, Y and Z
-components:
+We have been working with the direction of the magnetic field but what's its real magnitude?
+According to the documentation about the [`mag_data()`] function the x y z values we are
+getting are already in milli-gauss. That means the only thing we have to compute in order
+to get the magnitude of the magnetic field is the magnitude of the 3D vector that our x y z
```
eldruin, line 6:
```suggestion
According to the documentation about the [`mag_data()`] function the `x` `y` `z` values we are
getting are already in milli-gauss. That means the only thing we have to compute in order
to get the magnitude of the magnetic field is the magnitude of the 3D vector that our `x` `y` `z`
```

src/09-led-compass/magnitude.md
```diff
@@ -1,67 +1,85 @@
...
+values describe. As you might remember from school this is simply:
 
 ``` rust
-let magnitude = (x * x + y * y + z * z).sqrt();
+let magnitude = sqrtf(x * x + y * y + z * z);
```
eldruin, line 10: Maybe we should explain somewhere that we get use the `sqrt` function from `libm`?

src/09-led-compass/solution-1.md
```diff
@@ -51,21 +44,26 @@ fn main() -> ! {
...
-        while !sensor.mag_status().unwrap().xyz_new_data  {}
-        let data = sensor.mag_data().unwrap();
+        while !sensor.mag_status().unwrap().xyz_new_data {}
+        let mut data = sensor.mag_data().unwrap();
+        data = calibrated_measurement(data, &calibration);
 
         let dir = match (data.x > 0, data.y > 0) {
             // Quadrant I
-            (true, true) => Direction::SouthEast,
+            (true, true) => Direction::NorthEast,
```
eldruin, line 60: Signs still inverted? may be worth explaining the coordinate system?

eldruin (commented):
Looks good to me, thanks!

src/09-led-compass/solution-1.md
hargoniX: Yeah if I understood the microbit code that is being used for calibration and applying the calibration correctly they basically convert the values into a coordinate system where the pos. Y axis is north and the pos X axis is east. Like you would more or less intuitively assume considering the concept of north as "up" and east as "right".

hargoniX: My thought when writing this was basically that they should explore the coordinate system themselves (and since it looksl ike youd more or less intuitively assume it is easy to do so) but we can of course add a description.

eldruin: Ah, I see. We can leave it like this then.

eldruin:
Is this ready for merge now, then?

hargoniX:
We'd have to update the link to the lsm303agr docs that I use to justify the nT since its still saying mG but we can do that after your new minor release as well so basically yeah ready to merge and then I'll just have to clean it up once the release is there.

hargoniX:
Now its mergable.

eldruin (approved):
Thank you for all your work!
:tada: 

bors r+

PR #400, opened 2021-08-31
Use `cargo embed` instead of `cargo-embed`

caemor:
Replaces dash with whitespace since the version if the dash is not working.


eldruin (approved):
Good catch, thanks!
bors r+

PR #402, opened 2021-09-01
Add rustup target add to verification step and modify chapter 5 slightly

### LINKED ISSUE #401: [05-Rewrite] Led Roulette
In chapter 5 the necessary target gets added: https://github.com/rust-embedded/discovery/blob/rewrite/src/05-led-roulette/build-it.md but it is already necessary to add the target for the verification of the installation in chapter 3.
Should I make a PR moving everything upto line 33 to the verification chapter (next to the cargo embed stuff from #400) or is it better to just move the rustup target over?

caemor:
Adds `rustup target add *` to chapter 3 verification, but currently leaves the explanation in chapter 5 since the explanation fits better in there (it depends on explanations in ch4).
But if desired I can move everything around.

Fixes #401

### DIFF
--- a/src/03-setup/verify.md
+++ b/src/03-setup/verify.md
@@ -64,9 +64,11 @@ Next run one of these commands:
 ```
 $ # make sure you are in src/03-setup of the books source code
 $ # If you are working with micro:bit v2
+$ rustup target add thumbv7em-none-eabihf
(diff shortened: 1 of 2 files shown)


hargoniX (approved):
Thanks! bors r+

PR #403, opened 2021-09-11
Typos and grammar changes

robyoung:
Minor fixes and grammar changes on the micro:bit rewrite branch.

PR #404, opened 2021-09-11
Make terminology parts sub-sections

robyoung:
This allows us to deep link to them from other parts of the discovery book.

I've split this change out from #403 as it may be more contentious.


eldruin (approved):
Good point, thanks!
bors r+

PR #407, opened 2021-09-27
Rewrite polishing

hargoniX:


hargoniX:
I'm rather confused as to why this is blowing up during the linker stage because it can't find all these symbols. The micro:bit v2 version just works :tm: (at least locally).

eldruin:
hmm, no idea either. Have you tried to build the microbit v1 locally? Could it be a mixed v1/v2 flag/crate?

hargoniX:
Yeah it fails locally with the same linker error, the CI is not wrong here.

adamgreig:
I think this is most likely because two versions of cortex-m-rt are getting included, probably 0.6.14 and 0.7.0. You might find if you `cargo update` and re-run the build it fails with a more explicit error about not being able to resolve cortex-m-rt after https://github.com/rust-embedded/cortex-m-rt/issues/334. The solution is to either upgrade things using 0.6, or drop to 0.6 for whatever's on 0.7; you can't use both at once.

hargoniX:
Right now the build would be failing because of another reason anyways :(. nrf52388 hal made a release so their git repo is no longer version 0.13. However the microbit crate requires 0.13 at the moment (via git since I made some patched for this book to work) and hasn't updated to 0.14 yet because that would in turn break parts of microbit. So we will have to wait for this https://github.com/nrf-rs/microbit/pull/67 to land...until then the book build is unfortunately broken either way.

But I'll check your suggestion out once everything works again, thanks!

hargoniX:
Heureka the builds works \o/ @eldruin

eldruin (approved):
Awesome! Thank you very much for your work!
I am really looking forward to making the rewrite the main discovery book.

bors r+

adamgreig:
Nice! Is there anything left on the rewrite now besides merging this branch back in somehow so the new version is the main one?

hargoniX:
Not that I can think of no.

PR #418, opened 2021-12-15
Fix broken link to c3

caemor:
Change absolute link to relative since the current link is broken since it starts from the root level which is one above where it should be with the new structure.

### DIFF
--- a/microbit/src/05-led-roulette/debug-it.md
+++ b/microbit/src/05-led-roulette/debug-it.md
@@ -24,7 +24,7 @@ $ gdb target/thumbv6m-none-eabi/debug/led-roulette
 ```
 
 > **NOTE** Depending on which GDB you installed you will have to use a different command to launch it,
-> check out [chapter 3](/03-setup/README.md#tools) if you forgot which one it was.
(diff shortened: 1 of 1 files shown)


adamgreig:
Thanks! Weirdly the linkchecker didn't have a problem with it before, but now fails the CI with your correction. Maybe it's worth trying a relative URL without any preceding `..` or `/`?

caemor:
I tried to copy the behavior of https://github.com/rust-embedded/discovery/blob/master/microbit/src/05-led-roulette/debug-it.md?plain=1#L232 which is working.

In the CI the link-checker is currently run before the books are moved to their correct position and therefore the behavior might deviate from the expected behavior: https://github.com/rust-embedded/discovery/blob/4b5e5c73226c5a8d1565e2a65cbf2b709a0a231a/.github/workflows/ci.yml#L151

I think I now found the issue that stopped this link from working: https://github.com/rust-lang/mdBook/issues/984
Linking Readme.mds from other directories is broken. Workaround: Link to index.md which breaks the linkage outside of the website in your local folders when scrolling through the Markdownfiles

eldruin (approved):
Seems to work now, thanks!
bors r+

PR #422, opened 2021-12-30
Use specific chip for micro:bit v1 board.

gcapell:
probe-rs 0.12 changed from accepting generic chip description ("nrf51822")
to requiring specific chip description (e.g. "nrf51822_xxAA").

According to dirbaio on probe-rs matrix "there's three nrf51822 variants
with different flash/ram sizes, AA/AB/AC.
AA is 256k flash, 16k ram, afaik all microbit v1's are AA"

Without this change, when I attempted the `cargo embed` step I got
`Error... Found multiple chips matching 'nrf51822', unable to select a single chip.`

### DIFF
--- a/microbit/src/03-setup/verify.md
+++ b/microbit/src/03-setup/verify.md
@@ -53,7 +53,7 @@ chip variants:
 # v2
 # chip = "nrf52833"
 # v1
-# chip = "nrf51822-xxAA"
+# chip = "nrf51822_xxAA"
 ```
 
 If you are working with the micro:bit v2 board uncomment the first, for the v1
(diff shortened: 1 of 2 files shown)


microbit/src/03-setup/verify.md
```diff
@@ -53,7 +53,7 @@ chip variants:
 # v2
 # chip = "nrf52833"
 # v1
-# chip = "nrf51822"
+# chip = "nrf51822-xxAA"
```
Dirbaio, line 56: this is `_`, not `-`

gcapell: Oops. fixed with 657f0dd.

eldruin (approved):
Nice catch, thank you!
bors r+

hargoniX:
Doesn't this require change in basically all of the chapters? E.g. this one should blow up too https://github.com/rust-embedded/discovery/blob/master/microbit/src/05-led-roulette/Embed.toml

eldruin:
True. I prepared #423

PR #423, opened 2021-12-30
Fix micro:bit V1 chip specification everywhere

eldruin:
This is the fix from #422 propagated to all chapters.


adamgreig (approved):
thanks!

bors merge

PR #424, opened 2021-12-30
Update micro:bit chip descriptions in the book

eldruin:
I forgot editing the book as well in #423.

BTW, I put the chip instructions in one line each so that people are not confused and uncomment also the "v2" comment. I saw somebody asking in the chat about this.


adamgreig (approved):
bors merge

PR #426, opened 2021-12-31
In serial communication chapter, point to location of serial device f…

gcapell:
…or MacOS (info from https://makecode.microbit.org/device/serial).


eldruin (approved):
Nice improvement, thanks and happy new year!
bors r+

PR #427, opened 2022-01-03
Use more specific chip spec for micro:bit v2

### LINKED ISSUE #425: specify more specific chip for micro:bit v2?
Does a similar thing to #422 need to be applied for micro:bit v2?

eldruin:
Happy new year!

Fixes #425

### DIFF
--- a/microbit/src/03-setup/Embed.toml
+++ b/microbit/src/03-setup/Embed.toml
@@ -2,7 +2,7 @@
 protocol = "Swd"
 
 [default.general]
-# chip = "nrf52833" # uncomment this line for micro:bit V2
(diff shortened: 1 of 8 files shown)


adamgreig (approved):
:tada: 

bors merge

adamgreig:
Looks like the build failure is coming from a new 0.1.1 version of `nrf-usbd` which brings in `critical-section`, which [depends on std](https://github.com/embassy-rs/critical-section/blob/main/src/lib.rs#L110) when built for unix/windows, which brings in std's panic handler, which conflicts with `panic_rtt_target`/`panic_halt` from our own crates.

adamgreig:
bors retry

PR #430, opened 2022-01-11
Fixed "cargo size" argument

dobromyslov:
Fixed error: unknown argument '-version', did you mean '--version' when executing 'cargo size -- -version'.


adamgreig (approved):
Thanks! I think the syntax here changed since LLVM 11, but I expect most readers will have a version that supports `--version`: even LLVM 9 from rust 1.38 works with it.

bors merge

PR #431, opened 2022-01-12
Clarified 10.2 PuTTY configuration

dobromyslov:
Added new screenshow with PuTTY main screen and changed "Connection / Serial" screen from "As-is" to "To-be".


eldruin (commented):
Hmm, this removes the settings image as well, was that intentional? Showing the actual connection parameters is probably useful.

eldruin (approved):
Ah, I see it now, the image just got smaller in size. Sorry for the misunderstanding.
Looks good to me, thank you!
bors r+

PR #434, opened 2022-01-23
Update my-solutions in chapters 7 & 8 to working solutions (issues with heapless)

GreenPenguino:
This pull request adds the `my-solutions.md` of chapter 7 and 8 to work with version 0.7.x of heapless. 

Changes:
- Use `heapless = "0.7.10"` (because why not? I can leave it out if necessary)
- Use const generics with heapless (the `my-solutions.md` got left behind in #432)
- Use `core::str::from_utf8()` instead of `heapless::String::from_utf8()`, because the latter one [got removed](https://github.com/japaric/heapless/blob/3b2bc421a06c083dc71bd66c73f0d98f5976e0cb/CHANGELOG.md?plain=1#L123) in heapless `heapless = "0.7.x"`


adamgreig (approved):
Thank you, looks good to me!

bors r+

PR #443, opened 2022-02-09
Change minimal Rust toolchain version

AntoineRR:
This pull request updates the minimal Rust toolchain version required for microbit projects.

I had trouble compiling the microbit projects recently. After some research, I found out the `fixed` crate was updated to version `1.12.0` recently, and as stated on the [front page from the crate in crates.io](https://crates.io/crates/fixed), now requires Rust toolchain version 1.57.0 or higher.
This forces to have a Rust toolchain version >= 1.57.0 in order to compile the various projects for microbit v1 and v2.


eldruin (approved):
Thank you for looking into this and fixing it!
bors r+

PR #444, opened 2022-02-19
Fix wrong usage of atan2f

GreenPenguino:
The arguments of [atan2f](https://docs.rs/libm/latest/libm/fn.atan2f.html) are swapped. The first argument should be `y`, and the second should be `x`. This caused a mismatch between the axis in `take-1.md` and take-2.md`, where in take-1 the north is up, and in take-2 the north is on the right side. Also rotation was flipped because of this.

Swapping the arguments of `atan2f` to match the documentation solves the problem, but the if-else chain needed some adjustment as well.


eldruin (approved):
Nice catch, thanks!
cc: @hargoniX 
bors r+

PR #447, opened 2022-03-04
Update link to heapless::Vec to latest version

mathiasvr:


adamgreig:
Thanks for the PR! I wonder if using https://docs.rs/heapless/latest/heapless/struct.Vec.html to always point at the latest version would be even better? In theory it could break one day, but I expect heapless will probably have a `Vec` struct forever...

mathiasvr:
Yes makes sense, I updated the link to use latest version instead.

eldruin (approved):
Looks good to me, thanks!
bors r+

PR #454, opened 2022-05-29
Avoid pulling in microbit 0.13

shepmaster:


hargoniX:
If microbit already made a release upstream I don't see any reason to pull directly from git anymore, could we change it so that it uses crates.io like the other dependencies?

shepmaster:
Updated

hargoniX (approved):

hargoniX:
Looks good to me, thanks!
bors r+

therealprof:
What's so bad about 0.13?

hargoniX:
If 0.13 is fine we can pull it in as well, I just wanted to cut the pulling code in from Github in general.

therealprof:
Cool, I was just worried there was a problem with 0.13. I've heard people ran into problems using `defmt` 0.2 with new `probe-rs` tooling which was the main reason to remove `rubble` and update `defmt` in the latest `microbit`.

PR #456, opened 2022-06-28
Fix typo in microbit background's README.md

pokir:


eldruin (approved):
Good catch, thanks!
bors r+

PR #459, opened 2022-07-05
Add help tip for missing `cargo readobj` subcommand.

nutfinder2:
Adds install commands for `carg-binutils` if the reader does not already have it.

### DIFF
--- a/f3discovery/src/05-led-roulette/build-it.md
+++ b/f3discovery/src/05-led-roulette/build-it.md
@@ -123,4 +123,10 @@ ELF Header:
   Section header string table index: 20
   ```
 
+If you get an error like `error: no such subcommand: readobj`, you may need to install `cargo-binutils`:
+``` console
+cargo install cargo-binutils
+rustup component add llvm-tools-preview
+```
+
 Next, we'll flash the program into our microcontroller.


f3discovery/src/05-led-roulette/build-it.md
```diff
@@ -123,4 +123,10 @@ ELF Header:
   Section header string table index: 20
   ```
 
+If you get an error like `error: no such subcommand: readobj`, you may need to install `cargo-binutils`:
+``` console
+cargo install cargo-binutils
+rustup component add llvm-tools-preview
+```
+
```
eldruin, line 131:
```suggestion
> **NOTE** If you get an error like `error: no such subcommand: readobj`, you may need to install some packages. Please have a look at the [setup chapter](../03-setup/README.md#tools).
```

eldruin (commented):
Thanks for the suggestion!
However, installing cargo-binutils is already described in the [setup chapter](https://docs.rust-embedded.org/discovery/f3discovery/03-setup/index.html). Since not only this but also some other things need to be installed and set up in the machine, I would rather not add the duplicated information but add a link to the setup chapter.
Does that sound good to you?

PR #470, opened 2022-09-26
fix 05-led-roulette light-it-up.md

Jzow:
micro:bit v2 schematic image cannot be displayed normally because the github storage path has changed


eldruin (approved):
Thanks!
bors r+

Issue #479, opened 2022-11-11
Setup instruction does not work currently


Arthur-Milchior:
Hi,

I'm trying to follow https://docs.rust-embedded.org/discovery/microbit/03-setup/index.html
And I got stuck at `cargo install cargo-embed --vers 0.11.0`. Full error message below

I'm on ubuntu, did recent apt update/upgrade in case it matters. `rustc 1.65.0 (897e37553 2022-11-02)`. Other instructions above seems to work perfectly. At least I got no error message.

Thanks to ilyvion#9946 on discord I learned that I had to install libusb-1.0-0-dev (via apt on ubuntu). I'd suggest noting it somewhere in the documentation. It's indeed in  https://github.com/probe-rs/cargo-embed#linux so, you can also state to go read their documentation, even if it then become more complex

```shell
cargo install cargo-embed --vers 0.11.0
    Updating crates.io index
  Installing cargo-embed v0.11.0
   Compiling proc-macro2 v1.0.47
   Compiling quote v1.0.21
   Compiling unicode-ident v1.0.5
   Compiling syn v1.0.103
   Compiling version_check v0.9.4
   Compiling libc v0.2.137
   Compiling autocfg v1.1.0
   Compiling serde_derive v1.0.147
   Compiling serde v1.0.147
   Compiling cfg-if v1.0.0
   Compiling memchr v2.5.0
   Compiling log v0.4.17
   Compiling futures-core v0.3.25
   Compiling once_cell v1.16.0
   Compiling cc v1.0.76
   Compiling pkg-config v0.3.26
   Compiling ryu v1.0.11
   Compiling bitflags v1.3.2
   Compiling futures-io v0.3.25
   Compiling pin-project-lite v0.2.9
   Compiling thiserror v1.0.37
   Compiling linked-hash-map v0.5.6
   Compiling event-listener v2.5.3
   Compiling radium v0.5.3
   Compiling fastrand v1.8.0
   Compiling parking v2.0.0
   Compiling lazy_static v1.4.0
   Compiling waker-fn v1.1.0
   Compiling yaml-rust v0.4.5
   Compiling stable_deref_trait v1.2.0
   Compiling anyhow v1.0.66
   Compiling ahash v0.7.6
   Compiling value-bag v1.0.0-alpha.9
   Compiling crossbeam-utils v0.8.12
   Compiling cache-padded v1.2.0
   Compiling indexmap v1.9.1
   Compiling num-traits v0.2.15
[...]

PR #480, opened 2022-11-11
Add cargo-embed prerequisite

Arthur-Milchior:
Fixes: #479

Given that the prerequisites are relatively complex (depends on the OS), I preferred to link to them instead of trying to give specific instruction for all OSes. Hopefully, that should be enough for most learner, since the prerequisites section is pretty clearly marked

### DIFF
--- a/microbit/src/03-setup/README.md
+++ b/microbit/src/03-setup/README.md
@@ -76,7 +76,6 @@ cargo-embed 0.11.0
 git commit: crates.io
 ```
 
-
 ### This repository
 
 Since this book also contains some small Rust code bases used in various chapters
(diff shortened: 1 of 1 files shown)


microbit/src/03-setup/README.md
```diff
@@ -74,6 +76,7 @@ cargo-embed 0.11.0
 git commit: crates.io
 ```
 
+
```
eldruin, line 79:
```suggestion
```

eldruin (commented):
Looks good to me, thanks!

eldruin (approved):
bors r+

PR #486, opened 2022-11-25
Updating Chapter 8 with more inclusive terms

rmhyman:
The terms master/slave should not be propagated for future technologists to absorb. Made changes to text and images (some resizing may be necessary). I'm not sure this terminology is being used in subsequent chapters, but I'm more than happy to update them as I encounter them.

### DIFF
--- a/microbit/src/08-i2c/README.md
+++ b/microbit/src/08-i2c/README.md
@@ -22,7 +22,7 @@ used to synchronize the communication, this is a *synchronous* protocol.
 </p>
 
 This protocol uses a *controller* *target* model where the controller is the device that *starts* and
-drives the communication with a target device. Several devices, both conntrollers and targets, can be
+drives the communication with a target device. Several devices, both controllers and targets, can be
 connected to the same bus at the same time. A controller device can communicate with a specific target
 device by first broadcasting its *address* to the bus. This address can be 7 bits or 10 bits long.
 Once a controller has *started* a communication with a target, no other device can make use of the bus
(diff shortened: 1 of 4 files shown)


microbit/src/08-i2c/README.md
```diff
@@ -18,15 +18,15 @@ uses two lines to exchange data: a data line (SDA) and a clock line (SCL). Becau
...
 <p align="center">
-<img class="white_bg" height=180 title="I2C bus" src="https://upload.wikimedia.org/wikipedia/commons/3/3e/I2C.svg">
+<img class="white_bg" height=270 title="I2C bus" src="https://upload.wikimedia.org/wikipedia/commons/thumb/0/04/I2C_controller-target.svg/440px-I2C_controller-target.svg.png">
 </p>
 
-This protocol uses a *master* *slave* model where the master is the device that *starts* and
-drives the communication with a slave device. Several devices, both masters and slaves, can be
-connected to the same bus at the same time. A master device can communicate with a specific slave
+This protocol uses a *controller* *target* model where the controller is the device that *starts* and
+drives the communication with a target device. Several devices, both conntrollers and targets, can be
```
adamgreig, line 25:
```suggestion
drives the communication with a target device. Several devices, both controllers and targets, can be
```

adamgreig (changes requested):
Thanks! These are also the official terms for I2C now; the spec was updated last year.

I noticed one typo, and also I think it would be better if we could use the SVG version of the new images from Wikipedia - could you try changing to those please?

rmhyman:
> Thanks! These are also the official terms for I2C now; the spec was updated last year.
> 
> I noticed one typo, and also I think it would be better if we could use the SVG version of the new images from Wikipedia - could you try changing to those please?

That's good to know that the spec was updated 😃 . I've fixed the typo and updated the images to use the SVG file instead of the PNG version.

adamgreig (approved):
Thanks!

bors r+

PR #488, opened 2022-12-04
fix error code link

Jzow:


eldruin:
The HTML code is generated during the build so this will solve itself once #489 is through.
Thank you!
Closing.

Jzow:
@eldruin  Is this automatically generated based on the html code completed by github ci? `.md -> .html`

eldruin:
Yes, it is indeed.

PR #489, opened 2022-12-04
fix error code link

Jzow:


Jzow:
I don't think "here" should be used, but "code" can be used

eldruin (approved):
Thank you!
bors r+

PR #496, opened 2022-12-06
Fix incorrect html tag syntax

Jzow:
When I translate, the html syntax checks this error


eldruin (approved):
Nice catch, thanks!
bors r+

PR #497, opened 2022-12-06
Add code label

Jzow:
suggest code representation, The purpose is to unify.


eldruin (approved):
Sounds good to me, thanks!
bors r+

Issue #490, opened 2022-12-04
hidapi error: Permission denied


greenfierydragon:
When I run the following command:

```
$ cargo embed --features v1 --target thumbv6m-none-eabi
```

I get:

```
       Error Probe could not be created

             Caused by:
                 0: hidapi error: Permission denied
                 1: hidapi error: Permission denied
```

Despite having:

```
$ cat /etc/udev/rules.d/99-microbit.rules

# CMSIS-DAP for microbit
SUBSYSTEM=="usb", ATTR{idVendor}=="0d28", ATTR{idProduct}=="0204", MODE:="666"
```

I have rebooted several times and even checked the file permission on my device (`/dev/bus/usb/001/010`):

```
$ ls /dev/bus/usb/001/010
crw-rw-rw- 189,9 root  4 Dec 07:17 /dev/bus/usb/001/010
```

My kernel is 6.0.10 and I run archlinux. Any help would be much apreciated.

sudoBash418:
Not sure if this will fix your issue, but I was facing the same error and fixed it with the following udev rule:
```
# CMSIS-DAP for microbit
SUBSYSTEM=="usb", ATTRS{idVendor}=="0d28", ATTRS{idProduct}=="0204", MODE="0660", TAG+="uaccess"
```

According to the [Arch Wiki](https://wiki.archlinux.org/title/Udev#Allowing_regular_users_to_use_devices):
> The modern recommended approach for systemd systems is to use a `MODE` of `660` to let the group use the device, and then attach a TAG named `uaccess`.

After updating the udev rule, I ran `sudo udevadm control --reload; sudo udevadm trigger` and was then able to run `cargo embed --target thumbv6m-none-eabi` without issue.

Vortoman:
Had the exact same problem, solution worked. Thanks a lot!

eldruin:
Nice to see that you found a solution.
Could one of you put together a PR adding this to the book?

bcpeinhardt:
Same fix worked on Ubuntu, with a `udevadm trigger` instead of replugging in. :)

l-0-l:
Didn't work for me with a v1, not sure why. On Debian 12. Provided my solution [here](https://github.com/rust-embedded/discovery/issues/550).

BartMassey:
@l-0-l Thanks for that.

[...]

Issue #276, opened 2020-11-01
Linux install instructions include hacky, somewhat insecure, and often uneeded udev rule which makes mcu world-programmable


tim-seoss:
The Linux installation instructions include this sort of thing:

> These rules let you use USB devices like the micro:bit without root privilege, i.e. sudo.
> 
> Create this file in /etc/udev/rules.d with the content shown below.
> 
> $ cat /etc/udev/rules.d/99-microbit.rules
> 
> * CMSIS-DAP for microbit
> SUBSYSTEM=="usb", ATTR{idVendor}=="0d28", ATTR{idProduct}=="0204", MODE:="666"

This is a bit hacky, unnecessary on many Linux distros (e.g. Debian out of the box doesn't need this step I think, since the ACM dev get group writeable by the `plugdev` group) and also opens a bit of a security hole.  Although this is unlikely to be practically exploited, I think it should probably be fixed, since these sort of things have a habit of getting copy-pasted to things like embedded Linux systems, where they might end up doing some damage.

n.b. present on both master and rewrite branches

adamgreig:
I totally agree we shouldn't suggest a mode of 666 here; probably we should set the `uaccess` tag and `plugdev` group instead.

However, are you sure it's not required? For one thing, it's not a CDC-ACM device; it's either HID or bulk USB for the CMSIS-DAP interface used to program the board; but even with CDC-ACM I've usually found there does need to be a udev rule if it doesn't already recognise the device. However there are a _lot_ of extra rules hiding on most systems, for example on Ubuntu there's `/lib/udev/rules.d/60-openocd.rules` if you have `openocd` installed, which includes a rule for _all_ CMSIS-DAP devices to put them in the `plugdev` group and apply the `uaccess` tag, along with rules for almost all other probes openocd knows about.

shortspider:
I tested with Debian and Ubuntu and they both need rules defined for this to work.

What about setting the owner to the current user (`OWNER:="yourusername"`)?

PR #504, opened 2023-01-17
Fix issue #490: udev rule not working correctly on Arch Linux

sudoBash418:
As per issue #490, the current udev rule does not work correctly on Arch Linux.

This change is based on udev rules from [probe-rs](https://probe.rs/docs/getting-started/probe-setup/#udev-rules).

This new rule should work fine on any Linux distribution using systemd (or elogind), but I am unsure if it will also work on non-systemd distributions.

### DIFF
--- a/microbit/src/03-setup/linux.md
+++ b/microbit/src/03-setup/linux.md
@@ -60,12 +60,12 @@ These rules let you use USB devices like the micro:bit without root privilege, i
 Create this file in `/etc/udev/rules.d` with the content shown below.
 
 ``` console
-$ cat /etc/udev/rules.d/99-microbit.rules
+$ cat /etc/udev/rules.d/70-microbit.rules
 ```
 
 ``` text
 # CMSIS-DAP for microbit
-SUBSYSTEM=="usb", ATTR{idVendor}=="0d28", ATTR{idProduct}=="0204", MODE:="666"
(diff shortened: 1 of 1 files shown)


adamgreig (commented):
I wonder if it's worth leaving the mode at 666 or adding the plugdev group to increase support on systems that don't support the uaccess tag. Still, these days I think that's pretty much everything...

ddotthomas:
I was able to remove the 666 mode and use the plugdev group instead (which is also suggested at #276) by changing my rules file to look like this:
```
# CMSIS-DAP for microbit
SUBSYSTEM=="usb", ATTR{idVendor}=="0d28", ATTR{idProduct}=="0204", TAG+="uaccess", GROUP+="plugdev"
```

I'm using Pop OS however, I'm not sure if making the key SUBSYSTEMS and ATTRS is what fixes the issue over on Arch.

sudoBash418:
> I'm using Pop OS however, I'm not sure if making the key SUBSYSTEMS and ATTRS is what fixes the issue over on Arch.

I can confirm that using `SUBSYSTEM` and `ATTR` did *not* work for me on Arch Linux.

RabidGhost:
> I'm using Pop OS however, I'm not sure if making the key SUBSYSTEMS and ATTRS is what fixes the issue over on Arch.

Using SUBSYSTEMS and ATTRS, along with `TAG+="uaccess"` solved it for me on ubuntu.

PR #506, opened 2023-01-21
Factor out micro:bit LED roulette into turn-key example

sirhcel:
* Let's lower the bar four young players and have a working reference right out of the repository
* I could be put into action with `cargo-flash` with one of the following invocations:
    ```
    $ cargo flash --example my-solution --features v2 --target thumbv7em-none-eabihf --chip nrf52833_xxAA
    $ cargo flash --example my-solution --features v1 --target thumbv7em-none-eabihf --chip nrf51822_xxAA
    ```


eldruin (approved):
Good idea, thank you!
bors r+

PR #512, opened 2023-02-25
f3discovery: remove redundant cargo config from project directories

adicarlo:
Having a top level config config in f3discovery/src/.cargo/config.toml is a great idea, but the wonderful intentions are defeated by having per-project configs as well.

This PR removes those.

Note: microbit doesn't seem to have a shared cargo config, I left all that alone.


eldruin (approved):
Thank you for the clean up!
bors r+

PR #513, opened 2023-02-25
f3discovery: the gdb runner should reference our shared config

adicarlo:
Let `cargo run` be more seamless and automated in our shared F3discovery cargo configuration.

Make it clear that we expect OpenOCD to be running for this to work.  Yes, this is clarified in the book but a bit of redundant help might be welcome by newbies.


eldruin (approved):
Thank you!
bors r+

PR #521, opened 2023-04-22
Bump `cargo-embed` version in setup to 0.18.0

### LINKED ISSUE #515: Chapter 5.3: `monitor reset` is not supported by `cargo-embed` >= 0.13.0
Chapter 5 refers to `monitor reset` gdb command in the [`Debug it` section](https://github.com/rust-embedded/discovery/blob/7d879f301dc364b3858eff6fbb90514c50587e91/microbit/src/05-led-roulette/debug-it.md?plain=1#L190)

Unfortunately, `cargo-embed` (or rather `gdb-server`) dropped support for this command in 0.13.0 (https://github.com/probe-rs/probe-rs/issues/1564). Given that book [recommends installing `cargo-embed >= 0.11.0`](https://github.com/rust-embedded/discovery/blob/7d879f301dc364b3858eff6fbb90514c50587e91/microbit/src/03-setup/README.md?plain=1#L33), readers might benefit from a warning about this.

I've also had issues installing `cargo-embed` 0.11.0 or 0.12.0 due to dependencies being yanked, but this might be a user error :)

arr-ee:
Closes #515

I decided to not include a note about checking `cargo-embed` version in case `monitor reset` does not work, version range should be enough imo.

### DIFF
--- a/microbit/src/03-setup/README.md
+++ b/microbit/src/03-setup/README.md
@@ -30,7 +30,7 @@ should work but we have listed the version we have tested.
 
 [`cargo-binutils`]: https://github.com/rust-embedded/cargo-binutils
 
-- [`cargo-embed`]. Version 0.11.0 or newer.
+- [`cargo-embed`]. Version 0.18.0 or newer.
 
 [`cargo-embed`]: https://github.com/probe-rs/cargo-embed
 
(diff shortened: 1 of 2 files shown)


eldruin (approved):
Thank you for figuring this out and fixing it!
bors r+

PR #523, opened 2023-05-04
translation: add pt-BR link

allynaell:
Hi, I'm currently working on the Portuguese translation of this book and would like to add the link here. :)


adamgreig:
It looks like the link isn't working yet - could you get github pages building it first, so we don't add a dead link?

adamgreig (approved):
thanks very much, and thanks for the translation work!

bors r+

PR #525, opened 2023-05-12
Avoid world-writable devices in udev rules

mgeisler:
This came out of a discussion in Comprehensive Rust: https://github.com/google/comprehensive-rust/issues/608.

Basically, we've found that it's enough to make the device have `0664` permissions. If you agree with that, it would be nice to update the instructions here so that we can link to them from our course.

### DIFF
--- a/f3discovery/src/03-setup/linux.md
+++ b/f3discovery/src/03-setup/linux.md
@@ -142,7 +142,7 @@ sudo vi /etc/udev/rules.d/99-openocd.rules
 With the contents:
 ``` text
 # STM32F3DISCOVERY - ST-LINK/V2.1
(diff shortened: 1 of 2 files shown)


eldruin:
I principle I agree with you that it would be better to avoid making the devices world-writable, however, I worry about users that have a different or somewhat incorrect group configuration.
I would add more guidance here at least.
What do you think @adamgreig ?

adamgreig:
I'm not sure we need to even write 664 or just delete the MODE attribute entirely (I think 664 will be the default?). Removing world write access does make sense to me. Usually we'd either pug `GROUP="plugdev"` or `TAG+="uaccess"` in its place. I'm surprised that just 664 is enough without some other changes to ensure the user's in a group that can access the device or owns that device node. I don't think there's any harm having the plugdev group and uaccess tag added on systems that don't have them, either.

PR #529, opened 2023-05-25
chore: add vs-code settings for enabling cargo features

### LINKED ISSUE #528: Improvement: generate config for rust-analyzer
Was running into issues with rust-analyzer in VS Code with the whole repo open. Remembered dealing with this on the rustlings project, and they implemented this [lsp command](https://github.com/rust-lang/rustlings/blob/main/src/main.rs#L241) to generate a [rust-analyzer.json](https://github.com/rust-lang/rustlings/blob/main/src/project.rs). Rust analyzer supports [this](https://rust-analyzer.github.io/manual.html#non-cargo-based-projects) format.

margaretdax:
Created #528 before finding out the fix is much easier than I was imainging :)

Fixes #528

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "rust-analyzer.cargo.features": "all"
+}


eldruin (approved):
Looks good to me, thanks!
bors r+

PR #509, opened 2023-02-14
Fix link to cargo-embed prerequisites

90degs2infty:
Currently, the link pointing to `cargo-embed`'s prerequisites shows a redirection message: "cargo-embed has moved repository".

To ease the process of setting up the development environment, this PR replaces the outdated link by current a one. I'm not entirely sure if the updated link is the best one available, so feel free to substitute a more suitable one.

PR #535, opened 2023-07-17
Update cargo-embed links

icweaver:
Just continuing the dead link whac-a-mole game from #509. Happy to modify links/text if these are the wrong ones


eldruin (approved):
Looks good to me, thank you!

PR #556, opened 2024-02-15
Change target from thumbv7em-none-eabihf to thumbv7em-none-eabi

pythcoiner:
From [here](https://www.st.com/en/evaluation-tools/stm32f3discovery.html), it seems that Discovery F3 is a `Cortex-M4` not a `Cortex-M4F` so target should be `thumbv7em-none-eabi` instead `thumbv7em-none-eabihf` i guess.....

Not sure if its really matter, i cannot try on dicovery-F3, i only have Nucleo-F302R8 in hand and it seems to run fine w/ both targets...


jamesmunns:
Thank you, but the chip on that board does have an FPU: https://www.st.com/en/microcontrollers-microprocessors/stm32f303vc.html

"Cortex M4 with FPU" is the same as "Cortex M4F".

PR #571, opened 2024-06-08
Migrate to probe-rs-tools

nikgul:
Cargo-embed by it self has been deprecated, and has been included in probe-rs instead, update the documentation to install `probe-rs-tools` instead of `cargo-embed`, and added some command line tests to show how to test if probe-rs can see the micro:bit


eldruin (approved):
Looks good to me. Thank you for the update!

PR #572, opened 2024-06-08
Update udev rules 

nikgul:
Update udev rules to fit better with modern systems.

Updated the udev rules to use the `uaccess` tag instead of the `MODE` and `GROUP` modifiers, as that is deprecated.

Renamed the udev rule file to start with 69 instead of 99, so it gets loaded before user access rules are loaded for `uaccess` permissions, which is run in 73-seat-late.rules (provided by systemd).

Added text showing how to test for these `uaccess` permissions.

Attached the udev rules file: [69-microbit.txt](https://github.com/user-attachments/files/15748644/69-microbit.txt)

### DIFF
--- a/microbit/src/03-setup/linux.md
+++ b/microbit/src/03-setup/linux.md
@@ -60,21 +60,30 @@ These rules let you use USB devices like the micro:bit without root privilege, i
 Create this file in `/etc/udev/rules.d` with the content shown below.
 
 ``` console
-$ cat /etc/udev/rules.d/99-microbit.rules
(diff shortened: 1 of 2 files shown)


BartMassey (approved):
This looks great. Thanks much!

eldruin (approved):
Thank you!

nikgul:
Maybe this PR has fixed:
Issue: #490
PR: #504 & #525

And can therefore be closed ?

BartMassey:
It is possible. We should look into it.

Again, thanks.

PR #575, opened 2024-06-19
[WIP] Update the microbit part of the book

nikgul:
This is currently a work in progress.

Update the crate versions to the latest versions as I work my way through the book.

I plan on squasing the commits into one per chapter when this is ready.


microbit/src/05-led-roulette/debug-it.md
```diff
@@ -45,22 +45,23 @@ $ gdb target/thumbv6m-none-eabi/debug/led-roulette
 > as long as it does not crash, you are fine.
 
 Next we will have to connect to the GDB stub. It runs on `localhost:1337` per default so in order to
-connect to it run the following:
+connect to it run the following on the gdb commandline (`(gdb)`):
```
BartMassey, line 48: This is a bit confusing, I think. How about "on the `gdb` command line (that is, at the `(gdb)` prompt):"

microbit/src/05-led-roulette/Cargo.toml
```diff
@@ -2,28 +2,29 @@
 name = "led-roulette"
 version = "0.1.0"
 authors = ["Henrik Böving <hargonix@gmail.com>"]
-edition = "2018"
+edition = "2021"
 
 [dependencies.microbit-v2]
```
BartMassey, line 7: Most of the number corrections (version numbers, runtimes, entry points, sizes, etc) will change quickly as the ecosystem updates. A note to that effect should be put here somewhere, so that we aren't doing this on some regular basis. Probably should just freeze things at some point and make careful notes about the reference point, I guess? I don't know — it's a hard problem.

In a fantasy world, this kind of stuff would be done automatically in CI. But that sounds even harder.

BartMassey (changes requested):
Thanks much for this careful review. These changes mostly look fantastic. In particular, the use of `include` over pasting for the code examples is just great.

PR #577, opened 2024-07-06
add xHCI host dead troubleshooting

XMRHRX:
Hello, I want to contribute my solution about a problem trouble me a few months.

When STM32F freeze and not responding to Linux kernel, so kernel think it dead, reconnect USB won't make kernel re-check it.
I find the solution that we can use bind & unbind to make kernel re-check it.


BartMassey:
This sounds more like a Linux kernel bug than something the user should have to deal with? Maybe file a bug report there? (Yes, I know that reporting bugs on the Linux kernel is an adventure. But the host controller really shouldn't "get stuck" here, I think, which is what it looks like is happening offhand.)

XMRHRX:
Well, I think I need time to check these,
but I feel it less seem a bug, it feel more like this is just how it currently work, which if not get device response just remove it and never check it.
Maybe other "full implement" device like mouse, will make some request when re-connect, so kernel mark it connected again.(just a guess, I just dive into hardware recently)
Should I close this merge, or just keep it here till I have time to check it?

BartMassey:
Can leave it open for now. The messages look to me like the Linux host controller has gotten lost, but I could be wrong.

PR #582, opened 2024-09-23
 MacOS install instructions - lsusb

Theodlz:
I noticed while going through the discovery book (that uses a micro:bit controller), that I could not run the `lsusb` command provided after the setup section. It looks like on MacOS, this bin isn't there by default and also needs to be installed with `brew`, so this PR adds that command to the MacOS-specific installation steps.

### DIFF
--- a/microbit/src/03-setup/macos.md
+++ b/microbit/src/03-setup/macos.md
@@ -10,6 +10,9 @@ $ brew install arm-none-eabi-gdb
 
 $ # Minicom
 $ brew install minicom
+
+$ # lsusb (list connected USB devices)
+$ brew install lsusb
 ```
 
 That's all! Go to the [next section].


Theodlz:
I am aware that the section where `lsusb` is mentioned is flagged as `Linux` only, but I believe that having `lsusb` is super useful on MacOS even to just check that you are detecting the micro controller after plugging it in.

BartMassey (approved):
Thanks! This is especially useful on Mac, I think, where there are often problems getting started.

eldruin (approved):
Thank you!

BartMassey:
It would be great if someone with a Mac could also expand the Mac section to look more like the Linux section in terms of setting up and verifying setup of devices. I've filed an issue against

Theodlz:
@BartMassey I ordered the microbit v2 board and will get it next week. I will very likely open some PRs while I go through the guide, which might take me a few weeks (on vacation right now and traveling, so will get to that at best in a week, at most in 3 weeks).

BartMassey:
Thanks much! I've filed https://github.com/rust-embedded/discovery-mb2/issues/1 ; please do your review and PRs there, as this edition of the Discovery Book is being deprecated shortly.

PR #584, opened 2024-10-31
added instructions for finding Meta key on MacOS

8alpreet:
### The Issue
I had trouble using **minicom** because it's not obvious which keyboard key corresponds to **Meta.** I tried various keys such as CNTRL, CMD, & Option but nothing worked.

### Solution
I found that the Meta key is **ESC**  by running minicom in setup mode using `$ minicom -s`

### Changes
* I created a new page called `macos-meta.md`
* I added a link to it at the end of `nix-tooling.md`

### DIFF
--- a/microbit/src/06-serial-communication/minicom-on-macos.md
+++ b/microbit/src/06-serial-communication/minicom-on-macos.md
@@ -1,0 +1,10 @@
+# Minicom on macOS
+
+### Finding Your Command Key
+
+```shell
+$ # start minicom in setup mode
+$ minicom -s
+```
+1. Select `Keyboard and Misc Functions`
+2. Look at the first option `A - Command key is` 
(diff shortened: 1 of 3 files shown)


9names:
@therealprof what would you recommend instead?

BartMassey:
Please rebase this to current and I'll merge it. We're getting ready to close up this version of the Discovery Book and move to https://github.com/rust-embedded/discovery-mb2 . Bigger changes should happen there.

spookyvision:
I would probably recommend `pyserial-miniterm`. It's available through homebrew and has a command line option to customize the exit char (mine is mapped to Ctrl+X). Another option which comes bundled with MacOS is `screen` - mainly a terminal multiplexer but also has serial console capabilities.

BartMassey:
I have [a fork](https://github.com/BartMassey-upstream/scip) of [`scip`](https://gitlab.com/w0lff/scip) (AKA `serial-console`) that I've been using on a project. `scip` has the advantage that if you have Rust installed you can say `cargo install --locked serial-console` and you should be good to go cross-platform. My fork is probably not needed for Discovery Book usage, but I should probably try to get it upstreamed anyhow…

PR #585, opened 2024-11-08
ignore .idea everywhere

grodin:
Tiny QoL PR. 

There's no real reason (that I can see) to only ignore `.idea` in the repo root. If you only have a Microbit for example, you're going to work in the appropriate subdirectory and any `.idea` directory that gets created is just clutter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,4 @@
 Cargo.lock
 book/
 target/
-/.idea
+.idea


BartMassey:
Thanks for the suggestion!

Not sure git-ignoring an editor-specific dotfile is a great idea at all, actually: once we go down that road there's a lot of things we could do, but it probably belongs in folks' global gitignore instead.

I'm going to remove that line in https://github.com/rust-embedded/discovery-mb2 . Probably won't touch this, as we're trying to move over there soon.

Again, thanks.

grodin:
Yeah, that makes a lot of sense. I'll close this not to clutter up this repo.

PR #594, opened 2025-03-22
Mention the latest discovery book

Tomer-Eliahu:
Update the READMEs to mention the latest discovery book (https://docs.rust-embedded.org/discovery-mb2).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,14 +2,16 @@
 
 Discover the world of microcontrollers through [Rust](https://www.rust-lang.org/)!
 
-There are currently two versions of this book. Both of them provide an
+There are currently three versions of this book. All of them provide an
 introduction to microcontrollers and how to use Rust with them.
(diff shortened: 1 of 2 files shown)


BartMassey (changes requested):
Thank you very much for this! I should have done something long ago: the plan is actually to archive this repo fairly soon.

That said, I think we might be a little stronger about the language. I think we should be really clear that people should be using the new book at this point unless they really need to work with a micro:bit v1 or an STM32F3 Discovery Board. If you want to do this that's great; otherwise let me know and I'll take your PR and add to it later.

Again, really appreciate this PR.

Tomer-Eliahu:
I have made the changes to push people to use the new book. 
Is this what you had in mind? If not, please let me know what other fixes are needed.

BartMassey (approved):
This looks great! Thanks very much.

PR #597, opened 2025-04-07
fixed compiler warnings in snake-game

BartMassey:
This cleans up some old compiler warnings for the snake game.


eldruin:
The CI fails because it uses Ubuntu 20.04

jannic:
> The CI fails because it uses Ubuntu 20.04

I created https://github.com/rust-embedded/discovery/pull/598 to fix that

PR #599, opened 2025-12-16
terminology: crate `nrf51` has been renamed to `nrf51-pac`

LeoniePhiline:
See https://docs.rs/nrf51/0.10.0/nrf51/


BartMassey:
> Some people, among them me, have the discovery board v1. The new book does not cover the v1 at all. I very much hope that the original book can stay online.

It will stay online, just archived.

I'd encourage you to grab a v2 if possible though. It really is a better board, and the MB2 book has had many improvements.

PR #603, opened 2026-03-02
Docs microbit: replace archived link with current proyect link

lbellomo:
The books say: 

> high level crate already exists, like the [nrf52-hal](https://docs.rs/nrf52-hal)

but it's pointing to a mostly empty doc.
I replace that link for the correct one.


BartMassey (approved):
Thanks.

PR #607, opened 2026-03-20
remove links to imgur and replace with local file

### LINKED ISSUE #600: Replace imgur links
https://github.com/search?q=repo%3Arust-embedded%2Fdiscovery%20imgur&type=code

The search link above shows where it happens exactly.

One of the reasons is to prevent broken links (currently they serve "content unavailable", rather) for users accessing from the UK (sorry, that's bad).

Of course, one can use a VPN, but seems overkill in this case.

@jamesmunns 

See also: https://www.bbc.co.uk/news/articles/c4gzxv5gy3qo

lbellomo:
This PR fix #600, just add the git to the assert dir and replace the src in the links.
Just find 2 cases, but let me know if I miss another one.


BartMassey (approved):
Thanks!

PR #608, opened 2026-03-20
choose correct thumbnails size for mediawiki img

lbellomo:
I found a broken img to wikimedia and it's because the wiki try to standardize the size of the thrumbnails ([more info](https://www.mediawiki.org/wiki/Common_thumbnail_sizes) but TLDR: the 800px don't work but the next standard step 960px works).


BartMassey (approved):
Thanks!

